PR #6, opened 2015-05-27
Set auditor

ozzyjohnson:
Quick scripts to walk through the CF instance set auditing perms.

### DIFF
--- a/set-quota-auditor.sh
+++ b/set-quota-auditor.sh
@@ -13,10 +13,10 @@ cf orgs \
   | awk 'm;/^name/{m=1}' \
   | while read ORG_NAME
       do cf set-org-role $USER_NAME $ORG_NAME OrgAuditor
-         cf target -o $ORG_NAME
-  	     cf spaces \
-		  | awk 'm;/^name/{m=1}' \
-		  | while read SPACE_NAME
-		      do cf set-space-role $USER_NAME $ORG_NAME $SPACE_NAME SpaceAuditor
-		    done
+      cf target -o $ORG_NAME
+      cf spaces \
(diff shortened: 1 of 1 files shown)


set-quota-auditor.sh
```diff
@@ -0,0 +1,22 @@
...
+
+cf orgs \
+  | awk 'm;/^name/{m=1}' \
+  | while read ORG_NAME
+      do cf set-org-role $USER_NAME $ORG_NAME OrgAuditor
+      cf target -o $ORG_NAME
+      cf spaces \
+      | awk 'm;/^name/{m=1}' \
+      | while read SPACE_NAME
+          do cf set-space-role $USER_NAME $ORG_NAME $SPACE_NAME SpaceAuditor
```
afeld: Is this necessary? My understanding is that being an `OrgAuditor` cascades the permissions for the spaces beneath it.

afeld:
Pretty slick! Other than the one question, :ship:

set-quota-auditor.sh
ozzyjohnson: Yes, it seems to be necessary.

With `OrgAuditor` alone @ramirezg was getting back empty responses querying spaces for services, which are space-bound. 

I imagine it could be another case of the situation we had/have with `OrgManager` and `SpaceManager` where the backend knows how to enumerate permissions given the right API calls. 

We'd have to check which calls @ramirezg is making to see if there's are better / preferred endpoint(s) to the same end.

afeld: Ok cool we can figure that out separately.

PR #8, opened 2015-07-15
Add template change script showing the cf release template changes

dlapiduz:


dlapiduz:
@ozzyjohnson what do you think?

ozzyjohnson:
Looks fine. Only question I'd have is whether it would be worth the effort to add something to that would diff it the yml as an object.

PR #13, opened 2016-06-10
add a script for creating admins

afeld:


afeld:
Per the last part of https://docs.cloud.gov/ops/managing-users/#creating-admins, should these folks be made `OrgManager`s on the `cf` and/or `cloud-gov` orgs?

afeld:
@sharms Thoughts about the above?

sharms:
I am not sure, I have been admin privs so once that was set I never ran into a permissions issue again from a practical perspective.  Have you ran into that?

dlapiduz:
once we have admin we can make ourselves orgmanagers :)

PR #17, opened 2016-06-27
Script to create simple OAuth authentication UAA client

LinuxBozo:
Helps with 18F/cg-product#45

/cc @afeld

### DIFF
--- a/uaa-create-auth-client.sh
+++ b/uaa-create-auth-client.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+set -e
+
+if [ "$#" -ne 3 ]; then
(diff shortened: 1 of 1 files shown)


afeld:
## After merge

* [ ] Add link from https://docs.cloud.gov/apps/leveraging-authentication/

mogul:
I think rather than link from the user-facing docs, we should probably have this in the operator-facing docs located under "Contributing"... Operators are the ones who have to run the script.

rogeruiz:
Good point @mogul. @LinuxBozo and I talked about that during pairing and I see it as less than an hour's worth of work so I'm not reflecting it on the board. I'll link to this PR with the relevant PR from `cg-docs`.

PR #18, opened 2016-06-30
Add basic generate-passphrase script

rogeruiz:
This patch adds a script that helps standardize the way passphrases are generated to support the work on 18F/cg-docs#299.

### DIFF
--- a/generate-passphrase
+++ b/generate-passphrase
@@ -1,0 +1,6 @@
+#!/bin/sh
+LENGTH=32
+if [ ! -z "$1" ]; then
+  LENGTH=$1
+fi
+cat /dev/urandom | env LC_CTYPE=C tr -dc 'a-zA-Z0-9!@#$%^&*()_+?><~;' | fold -w $LENGTH | head -n 1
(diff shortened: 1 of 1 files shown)


generate-passphrase
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+LENGTH=32
+if [ ! -z "$1" ]; then
+  LENGTH=$1
+fi
+cat /dev/urandom | env LC_CTYPE=C tr -dc 'a-zA-Z0-9!@#$%^&*()_+?><~;' | fold -w $LENGTH | head -n 1
```
LinuxBozo: @rogeruiz You might find this doesn't work too well in OSX. At least, not for me in Yosemite.. However, I have a version which does work..

```
cat /dev/urandom | LC_ALL=C tr -dc 'a-zA-Z0-9!@#\$%^&\*\(\)\+=_' | head -c $LENGTH
```

afeld: Ah, that's better than what I provided! Also doesn't have a trailing newline, which is good.

afeld: Now mine's all tricked out: https://github.com/afeld/dotfiles/blob/master/bin/generate-password 🚗  💨

PR #19, opened 2016-07-01
Add ignorance to the project

rogeruiz:
With the work being done in 18F/cg-docs#299 around automating uaac client
authorization creation, the docs suggest creating `*-auth` files to help
with the transfer of credentials after it's created. The project didn't
have any .gitignore values so this patch adds them and helps with the
workflow mentioned above.


afeld:
Ignored!

PR #29, opened 2016-11-22
refactor Ruby scripts

afeld:
* Various formatting cleanup, including fixing indentation to use spaces rather than tabs, which is a convention for Ruby.
* Share the CF CLI logic between the scripts.
* Fix incorrect file extension.

Note: I did not validate all of the scripts by hand.


cf-servicesinfo.rb
```diff
@@ -3,107 +3,96 @@
...
-		end
-		page += 1
-	end
-
-	# reduce the service plans hash to only those plans that have active instances, and sort them
-	# in descenting order
-	service_plans.delete_if {| k, v | v['service_instance_count'] == 0 }
-	service_plans = service_plans.sort_by { |k, v| v['service_instance_count'].to_i }.reverse
+  page = 1
+  while page < 3
```
jcscottiii: any reason why we only look at the first two pages? not a blocker
(I know nothing changed but this was just shifting the indents)

afeld: @jbarnicle?

jbarnicle: That is a good catch @jcscottiii .  I knew (at the time) that that there were only two pages so I lazily hardcoded this.  It should be coded as the other paged results are - along the lines of (I haven't checked this for syntax, etc)
```
page = 1
more_service_instances = true
while more_service_instances
   service_instances =  cfmunge("/v2/service_instances?page=#{page}&results-per-page=100")
   if service_instances.length > 0
       service_instances.each do |service_instance|
           # do something
       end
       page += 1
   else
        more_service_instances = false
   end
end

PR #32, opened 2016-11-25
refactor the services info script

afeld:
Some random cleanup, most importantly: centralizing the API pagination logic, as a generator.

/cc @datn @jcscottiii

### DIFF
--- a/cf-servicesinfo.rb
+++ b/cf-servicesinfo.rb
@@ -10,16 +10,7 @@ require_relative 'helpers'
 
 
 def exempt_space(space_name)
-  if space_name.include?('stag') ||
-    space_name.include?('dev') ||
-    space_name.include?('test') ||
-    space_name.include?('sandbox') ||
-    space_name == 'alan' ||
-    space_name == 'yoz'
-    return true
-  end
-
-  return false
+  space_name =~ /stag|dev|test|sandbox|\A(alan|yoz)\z/
 end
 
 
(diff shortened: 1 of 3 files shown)


cf-servicesinfo.rb
```diff
@@ -163,23 +136,25 @@ def get_space_service_instances(space_guid, service_plans)
 stamp = Time.now.to_i
 
 CSV.open("cf-orgs-services-managers-#{stamp}.csv", "wb") do |csv|
-  csv << ["Org","Org Managers","Space", "Space Developers", "Space Developers", "Space GUID"] + get_service_plan_headers(@service_plans)
+  csv << ["Org","Org Managers","Space", "Space Developers", "Space Developers"] + get_service_plan_headers(@service_plans)
```
jcscottiii: this should be "Org","Org Managers","Space", **"Space Managers"**, "Space Developers"

cf-servicesinfo.rb
```diff
@@ -10,38 +10,22 @@
 
 
 def exempt_space(space_name)
```
jcscottiii: can we change this to `exempt_space?` or `is_exempt_space/is_exempt_space?`. just so it's easier to know that it's meant to be returning true or false

cf-servicesinfo.rb
jcscottiii: bump on this one

Issue #31, opened 2016-11-23
use new quota naming structure


afeld:
Per @berndverst [in Slack](https://gsa-tts.slack.com/archives/cloud-gov-skyporter/p1479932190004143):

> For the quota naming, I’ve been using: `AGENCY-NAME_IAA-NUMBER_POPSTART-POPEND-01`

...where `POP` is "period of performance".

berndverst:
e.g. `20161101-20171031-01` where 01 is a counter. If this quota already exists, increment the quota by one and create that instead.

PR #26, opened 2016-11-09
add validation to org creation script

afeld:
Paired with @jcscottiii.

PR #34, opened 2016-11-25
replace the NOTE with the period of performance when creating orgs

afeld:
Closes https://github.com/18F/cg-scripts/issues/31. Builds on https://github.com/18F/cg-scripts/pull/26.

/cc @jcscottiii @berndverst

### DIFF
--- a/cf-create-org.sh
+++ b/cf-create-org.sh
@@ -3,17 +3,18 @@
 set -e
 
 if [ "$#" -lt 5 ]; then
-  printf "Usage:\n\n\t\$./cf-create-org.sh <AGENCY_NAME> <IAA_NUMBER> <SYSTEM_NAME> <NOTE> <MANAGER> <MEMORY>\n\n"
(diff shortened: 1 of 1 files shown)


cf-create-org.sh
```diff
@@ -30,17 +31,22 @@ if ! [[ $SYSTEM_NAME =~ ^[a-zA-Z0-9-]+$ ]]; then
...
   exit 1
 fi
 
-QUOTA_NAME="${AGENCY_NAME}_${IAA_NUMBER}_${NOTE}"
+if ! [[ $POP_END =~ ^[0-9]{8}$ ]]; then
+  echo "POP_END should be of the format YYYYMMDD."
+  exit 1
+fi
+
+QUOTA_NAME="${AGENCY_NAME}_${IAA_NUMBER}_${POP_START}-${POP_END}-01"
```
jcscottiii: do we want to handle the auto-increment of the counter in this PR? 
ref: https://github.com/18F/cg-scripts/issues/31#issuecomment-262622057

afeld: Ummm, not sure how to handle that cleanly, so would like to punt on it if that's ok with you. We can open an issue...?

PR #43, opened 2017-02-16
Add change quota plan script

rogeruiz:
I had to change some quotas for three different orgs, so I made this little script to help.


rogeruiz:
Closing since I don't know how to read the docs. You could boil the above script down to 😁 

```
cf set-quota $ORG $NEW_QUOTA
```

PR #44, opened 2017-02-17
Bump default number of routes and services to 20

brittag:
10 is too low for many, and it doesn't really cost us anything to have a bigger default.


brittag:
Closing this since completed at https://github.com/18F/cg-scripts/pull/45

PR #46, opened 2017-03-13
Force a stuck cdn broker instance to be provisioned.

jmcarp:
Until the service broker api gets the concept of cancelling an in-progress instance.

### DIFF
--- a/poke-cdn.sh
+++ b/poke-cdn.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# Force a cdn broker instance into the `provisioned` state so that it can be deleted
+# Usage: `DOMAIN="some.domain.gov" ./poke-cdn.sh`
+
+set -e
+set -u
+set -x
+
+if ! cf plugins | grep "connect-to-service"; then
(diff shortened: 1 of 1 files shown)


poke-cdn.sh
```diff
@@ -0,0 +1,38 @@
...
+
+if ! cf plugins | grep "connect-to-service"; then
+  echo "connect-to-service plugin not found; install at https://github.com/18F/cf-service-connect#local-installation"
+  exit 1
+fi
+
+creds="creds.out"
+
+cf target -o ${ORG:-cloud-gov} -s ${SPACE:-services}
+cf connect-to-service --no-client ${APP:-cdn-broker} ${DATABASE:-rds-cdn-broker} > ${creds} &
```
cnelson: Can the script from this point on be simplified to just this oneliner?

`echo "update routes set state = 'provisioned' where domain_external = '${DOMAIN}' and state = 'provisioning';" | cf connect-to-service ${APP:-cdn-broker} ${DATABASE:-rds-cdn-broker} > ${creds}`

jmcarp: Yep, amended.

PR #48, opened 2017-04-19
Add in recent users script

rogeruiz:
This quick script should help for getting a rudimentary CSV list of recent users from the platform.

cc: @jameshupp


rogeruiz:
This is still a WIP until we get a better `jq` filter for removing `deployer-service` accounts. Also, this is useful only to users with `cloud_controller.admin` so @jcscottiii @fureigh @rememberlenny should be able to pull this data once the script is updated.

cnelson:
You can bypass the CF API and get a list of users from UAA directly.  This can be done with a lesser scope (`scim.read`) which we could probably grant to folks who need to run this.

```python
#!/usr/bin/env python

import argparse
import csv
import datetime
from email.utils import parseaddr
import subprocess
import sys
import yaml

def make_date(datestr):
    return datetime.datetime(*[int(x) for x in datestr.split('-')])

if __name__ == '__main__':
    parser = argparse.ArgumentParser(description="Show a list of users in UAA.")
    parser.add_argument('since', help="Return users created on or after this date YYYY-MM-DD", type=make_date)

    args = parser.parse_args()

    try:
        check = yaml.load(subprocess.check_output(
            ['uaac', 'users', '--attributes', 'username', '--count', '1']
        ))
        total = check['totalresults']
        data = yaml.load(subprocess.check_output(
            ['uaac', 'users', '--attributes', 'username', '--attributes', 'meta.created', '--attributes', 'verified', '--count', str(total)]
        ))
        assert data['itemsperpage'] == total
    except subprocess.CalledProcessError as exc:
        parser.error("Unable to list UAA users:{}\nRequest a token with: uaac token sso get cf -s '' --scope scim.read".format(exc.output))
    except AssertionError:
        parser.error("UAA returned a truncated response.  Time to implement paging :(")

writer = csv.DictWriter(sys.stdout, fieldnames=['username', 'meta.created'], extrasaction='ignore', dialect='excel-tab')
writer.writeheader()
for user in data['resources']:
    if '@' in parseaddr(user['username'])[1] and user['meta.created'] >= args.since:
        writer.writerow(user)
```

rogeruiz:
Still leaving this as WIP until there are better docs around how to set this up locally if you're not a :cloud:.gov operator, but once that's done we should be good to have almost anyone on the team pull this data from UAA. cc @gboone

cnelson:
@rogeruiz I think it's just a matter of folks who need access asking Cloud Operations to grant them `scim.read` scope + installing and configuring uaac which has some OK docs in it's readme: https://github.com/cloudfoundry/cf-uaac. 

TL;DR
```
gem install cf-uaac
uaac target uaa.fr.cloud.gov
uaac token sso get cf -s '' --scope scim.read
python cf-get-recent-users.py YYYY-MM-DD
```

gboone:
Two things from taking a stab at this:
1. The passcode prompt from `uaac token sso get cf -s '' --scope scim.read` says to grab the token from https://uaa.fr.cloud.gov/passcode which redirects to https://login.fr.cloud.gov. Don't know what the easiest solution would be but https://login.fr.cloud.gov/passcode/ got me set.
2. The python command fails with an import error `ImportError: No module named yaml`.

gboone:
I can't comment on the code, here, except to say it works. I _can_ comment on the docs in the README and they look 💯 Happy to make that a formal review and unblock merging if that's helpful.

brittag:
This should also have a code review before merge. :)

PR #51, opened 2017-05-10
Filter Repos By Ownership

commit-dkp:
To manage security, we need to understand where our internal processes
and external dependencies are.  This script extracts all of the github
URIs from ConcourseCI pipelines and then buckets them into repos that
are under the 18F organization and repos that are not.


jmcarp:
Works as advertised. Thanks for the chrome @cnelson !

PR #60, opened 2017-10-27
Add cg-common resources grabber

rogeruiz:


jmcarp:
Deprecating cg-common.

PR #61, opened 2017-11-03
Add simple parser for XML ZAP report for cross ref to POAM

LinuxBozo:
This makes sure that we've stripped anything that doesn't match cloud.gov (like google analytics bits) that may get picked up in the scan spidering process.

example output:
```
SQL Injection, CWE id 89, WASC id 19
	https://foo.cloud.gov
	https://bar.cloud.gov
X-Frame-Options Header Not Set, CWE id 16, WASC id 15
	https://bar.cloud.gov
```

POAM sheet should contain the CWE and WASC ids as `Weakness Reference Identifiers` from previous ZAP reports for cross reference.

/cc @brittag

### DIFF
--- a/parse-owasp-zap-xml.py
+++ b/parse-owasp-zap-xml.py
@@ -1,0 +1,26 @@
+import xml.etree.ElementTree as etree
+import sys
+
+if len(sys.argv) == 1:
+    print('please provide a path to an XML ZAP report')
+    sys.exit(-1)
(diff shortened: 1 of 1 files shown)


brittag:
Yay! Might make sense to cross-link this at a useful spot in https://github.com/18F/cg-product/blob/master/ConMonChecklist.md to remind us we can run it to help with processing.

PR #63, opened 2017-11-14
[WIP] Automate fetching and combining secrets for use in concourse pipelines

wjwoodson:
Add `generate-concourse-environment.sh` for operators to generate a concourse env file `concourse-environment.yml` (`--load-vars-from`) at run time when updating pipelines.

`CI_ENV=concourse_target SECRETS_BUCKET=s3_bucket CG_PIPELINE=~/cg-pipeline-tasks ./generate-concourse-environment.sh cg-vars-file.yml`

Requires:
- operator access to concourse pipelines (secret-rotation)
- read access to encrypted s3 SECRETS_BUCKET
- a local copy of https://github.com/18F/cg-pipeline-tasks (decrypt)

Optional:
- any number of `cg-vars-file.yml` to populate pipeline specific, non-secret vars.

Goes with work in https://github.com/18F/cg-secret-rotation/pull/6 and https://github.com/18F/cg-deploy-bosh/pull/179

### DIFF
--- a/generate-concourse-environment.sh
+++ b/generate-concourse-environment.sh
@@ -1,0 +1,95 @@
+#!/bin/bash
+
+set -eu
+
+RED='\033[0;31m'
+CYAN='\033[0;36m'
+GREEN='\033[0;32m'
+YELLOW='\033[0;33m'
+PURPLE='\033[0;35m'
+NC='\033[0m'
+
+if [[ -z $CG_PIPELINE ]]
+then
+  echo -e "${RED}ERROR${NC} Please set a ${YELLOW}\$CG_PIPELINE${NC} variable pointing to a clone of ${YELLOW}https://github.com/18F/cg-pipeline-tasks${NC}"
+  echo -e "eg, ${PURPLE}CG_PIPELINE=~/dev/cg-pipeline-tasks ./generate-k8s-release-certs.sh"
+  exit 98
+fi
+
+if [[ -z $SECRETS_BUCKET ]]
+then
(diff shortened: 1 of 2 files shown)


generate-concourse-environment.sh
```diff
@@ -0,0 +1,95 @@
...
+  echo -e "${RED}ERROR${NC} Please set a ${YELLOW}\$CI_ENV${NC} with the name of the ${YELLOW}concourse target where the secret-rotationpipeline is stored${NC}"
+  echo -e "eg, ${PURPLE}CG_PIPELINE=~/dev/cg-pipeline-tasks ./generate-k8s-release-certs.sh"
+  exit 98
+fi
+
+# create a combined secrets file
+echo "foo: bar" > secrets-combined.yml
+
+# get environment secrets files
+for ENVIRONMENT in $(echo 'common tooling development staging production' | tr " " "\n")
```
wjwoodson: `common` needs to become `master`

generate-concourse-environment.sh
```diff
@@ -0,0 +1,95 @@
...
+
+if [[ -z $CI_ENV ]]
+then
+  echo -e "${RED}ERROR${NC} Please set a ${YELLOW}\$CI_ENV${NC} with the name of the ${YELLOW}concourse target where the secret-rotationpipeline is stored${NC}"
+  echo -e "eg, ${PURPLE}CG_PIPELINE=~/dev/cg-pipeline-tasks ./generate-k8s-release-certs.sh"
+  exit 98
+fi
+
+# create a combined secrets file
+echo "foo: bar" > secrets-combined.yml
```
LinuxBozo: Can we make this initial key a bit more random so we can get rid of any race condition where foo might be a valid key?  Also, this will make sure that if valid later, it doesn't get pruned in the later merge.

PR #64, opened 2017-11-16
Obfuscate the quota name + delay fly check

mogul:
* To avoid quotas revealing info about other customers in the system, we create it under a hashed name; the relationship between orgs and quotas is what we care about, not the quota names!
* To make testing easier, don't check that we're logged into Fly until we're actually ready to go through with it!

### DIFF
--- a/cf-create-org.sh
+++ b/cf-create-org.sh
@@ -16,14 +16,6 @@ MANAGER=$6
 
 MEMORY="${7:-4G}"
 
-CI_URL="${CI_URL:-"https://ci.fr.cloud.gov"}"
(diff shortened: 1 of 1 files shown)


mogul:
Out of band: I've [manually renamed all the existing "revealing" quota names to their corresponding md5sum](https://gsa-tts.slack.com/archives/C2Q8LMFNH/p1510830988000048).

cc @brittag

brittag:
@sharms Would you be up for reviewing this? This would still be a helpful improvement for the org creation script. Thank you!

sharms:
Looks good to me!

PR #72, opened 2018-02-28
Check that DOMAIN is a valid domain name, doesn't have spaces etc

sharms:


poke-cdn.sh
```diff
@@ -12,6 +12,11 @@ if ! cf plugins | grep "connect-to-service"; then
   exit 1
 fi
 
+if [[ $(echo "$DOMAIN" | pcregrep '(?=^.{5,254}$)(^(?:(?!\d+\.)[a-zA-Z0-9_\-]{1,63}\.?)+\.(?:[a-z]{2,})$)' | wc -l) -ne 1 ]]; then
```
jmcarp: What if `domain_external` includes multiple domains, like `my.domain.gov,my-other.domain.gov`?

bengerman13:
closing since poke-cdn is no longer used

PR #73, opened 2018-06-29
added cloudtrailcheck script

timothy-spencer:


cloudtrail-check.sh
```diff
@@ -0,0 +1,55 @@
...
+# 
+# This script is here to automate (as much as possible) the tasks in
+# https://cloud.gov/docs/ops/maintenance-list/#review-aws-cloudtrail-events
+# 
+# authorized events will include:
+#   a user name like i-17deadbeef1234567
+#   a source IP address within AWS.
+#   an AWS access key starting with ASIA
+# 
+# XXX The source IP range is kind of janky.  It would be better if this
```
jmcarp: How janky? What specifically is wrong with it? Can we do handle addresses better?

cloudtrail-check.sh
```diff
@@ -0,0 +1,55 @@
...
+#     pulled dynamically from AWS and also if we handled cidr notation
+#     properly.  But these two ranges seem to work for govcloud right now.
+
+START=$(date -v -15d +"%Y-%m-%dT%H:%M:%SZ")
+echo "pulling cloudtrail from a bit over 2 weeks ago ($START)"
+
+aws cloudtrail lookup-events --start-time "$START" --lookup-attributes AttributeKey=EventName,AttributeValue=ConsoleLogin > /tmp/consolefoo.$$
```
jmcarp: Does this handle pagination?

cloudtrail-check.sh
```diff
@@ -0,0 +1,55 @@
...
+#   a user name like i-17deadbeef1234567
+#   a source IP address within AWS.
+#   an AWS access key starting with ASIA
+# 
+# XXX The source IP range is kind of janky.  It would be better if this
+#     pulled dynamically from AWS and also if we handled cidr notation
+#     properly.  But these two ranges seem to work for govcloud right now.
+
+START=$(date -v -15d +"%Y-%m-%dT%H:%M:%SZ")
+echo "pulling cloudtrail from a bit over 2 weeks ago ($START)"
```
jmcarp: Let's be precise about the date range here.

jmcarp:
Could we automate this further? What if it ran on a timer in concourse and pinged slack or alertmanager if operator attention is required?

cloudtrail-check.sh
timothy-spencer: I am.  I expect that there will be some fuzz factor for when the operator runs this, so if we just do 2 weeks, we risk not examining data if one operator ran it at 9am on monday of their first shift, and the next operator runs it at 10am on monday of their first shift.  I figured that 1 day of fuzz factor would be sufficient.

cloudtrail-check.sh
timothy-spencer: I didn't experience any pagination while running this, but that's definitely something that I will add.  Thanks!

cloudtrail-check.sh
timothy-spencer: It is janky because it's not handling CIDR notation for realz.  So if you have a 10.10.0.0/17, this is a PITA because you'd have to enumerate 10.10.{0-127}.0 for matching.  It is possible that this could be done with some regex/jq gymnastics, but I couldn't think of a great approach, so I just punted and went with classful matching.

timothy-spencer:
@jmcarp Yeah, it certainly could be plugged into some sort of job and alerted on.  I just cooked this up while doing https://cloud.gov/docs/ops/maintenance-list/#review-aws-cloudtrail-events and thought that it would be better than what we had right now (no automation at all).

When I have time, I'd be happy to do more.

cloudtrail-check.sh
rogeruiz: @timothy-spencer do we need to handle pagination here? I'm confused.

rogeruiz:
Hey, @timothy-spencer I'd like a demo of this and how it works. thanks!

cloudtrail-check.sh
LinuxBozo: I think what @jmcarp is getting at is, if it's 15 days, say 15 days, not "a bit over 2 weeks ago"

timothy-spencer:
@LinuxBozo I wrote the script so that it would be easy to show the output to an auditor and be able to say "look, the audit script said that nothing should be here, and nothing was here, so we are good".  In my experience, this is a feature that they appreciate for their screenshotting.  I'll try to rework it so that we can have both types of views.

Good point on understanding the first section, and thanks for clarifying what Josh meant.  I'll clean those up.

rogeruiz:
After running this locally, I don't really like how this runs and exits with a `non-zero` status and keeps stuff in my `stdout`. I think keeping the file around or piping this through `less` for output would be better. But it's _good enough_ for now. I think the output from AWS is too tall for my screen so I don't get to see *investigating*. I'm on a 13" laptop.

timothy-spencer:
I would hope that operators could see that it was larger than the screen when running the script and use `less` or `more` or shell scrollback or a redirect to a file and open with an editor or whatever they wanted to view the output of the script.  I'm not sure what change you are looking for here.  Handling longer command output is a pretty basic thing that all cloud.gov operators should be familiar with.

The non-zero status indicates that audit records were found, so that this script could be later on used in a pipeline.  There is no downside to non-zero return values, especially if you are just running it from the commandline.  Again, I'm not sure what you are looking for here.  This doesn't break anything, and it is a good future-proofing thing to have.

rogeruiz:
> hey @sharms & @timothy-spencer,

> Thanks for bringing this up @sharms. We shouldn't have bad/wrong documentation, especially for support/interrupt when we have folks new to it working on it for the first time. Let's move fast and change things by iterating on them properly. @timothy-spencer your work isn't being removed from 18F/cg-scripts with this work. It's only the documentation that points to it.

> As the person who merged this in originally, along with the original script as well, I think it's a good idea to step back and refocus these efforts. @sharms has mentioned a few times here that the documentation is clearly lacking. Thank you @timothy-spencer for the effort of improving a process which surfaced these deficiencies in the current process. @sharms mentions that an operator should be able to see any issues which the manual review accomplishes by having operators comb through logs. I see he also thinks that having your script alongside the manual process really is the best of both worlds here.

> Let's add the automation to Concourse here to make things easier and I propose we break apart your script into a group of scripts that do one thing really well. That way we can take the most value out of your work and fix the issues that came up here in the other parts.

> I'm going to approve this PR, and we can bring this conversation over to the original PR adding the script to improve it. 18F/cg-scripts#73

~ source https://github.com/18F/cg-site/pull/1320#issuecomment-411795454

Let's continue the conversation here pointing out stuff we can split out. I can start.

cloudtrail-check.sh
```diff
@@ -0,0 +1,81 @@
...
+
+START=$(date -v -15d +"%Y-%m-%dT%H:%M:%SZ")
+echo "pulling cloudtrail from a 15 days ago so that 2 week audits will always overlap a bit ($START)"
+
+aws cloudtrail lookup-events --no-paginate --start-time "$START" --lookup-attributes AttributeKey=EventName,AttributeValue=ConsoleLogin > /tmp/consolefoo.$$
+
+echo "============ Console logins per user"
+echo "============ Run 'aws cloudtrail lookup-events --no-paginate --lookup-attributes AttributeKey=EventName,AttributeValue=ConsoleLogin'"
+echo "============ for details on the logins if something seems unusual about the users or the number of times they logged in."
+cat /tmp/consolefoo.$$ | jq -r .Events[].Username | sort | uniq -c
```
rogeruiz: These couple of lines about `Console logins per user` should be its own script.

cloudtrail-check.sh
```diff
@@ -0,0 +1,81 @@
...
+echo
+
+cat /tmp/consolefoo.$$ | jq -r '.Events[] | select(.Username == "Administrator")' > /tmp/admin.$$
+cat /tmp/consolefoo.$$ | jq -r '.Events[] | select(.Username != "..*") | .CloudTrailEvent' | \
+	jq -r 'select(.userIdentity.userName == "Administrator")' >> /tmp/admin.$$
```
rogeruiz: This should be it's own script as well.

cloudtrail-check.sh
rogeruiz: @LinuxBozo I believe @timothy-spencer original idea with this is taken from this from the docs.

> These EventNames should be attributed to human individuals on the cloud.gov team:
> - `ConsoleLogin`

The idea here is that if we know someone is out of the office, or has had access removed, they should no longer have `ConsoleLogin` appears any more. So it's to be expected that their access should be at `zero` if they're not around. Does that make sense?

I think the value add isn't as high as some of the other metrics we look for while on support, but having it in a pipeline running every ten minutes or every half hour would be more helpful than ignoring the metric completely.

rogeruiz:
@timothy-spencer adding a pipeline with a task that runs an arbitrary shells script is [described in the Concourse docs here](https://concourse-ci.org/tasks.html) and there's [an example I did recently here](https://github.com/18F/cg-deploy-stratos/pull/2/files).

sharms:
The current behavior of this script's exit code:

1. When ran without the AWS CLI installed or `jq`, it exits success
1. When ran without internet connectivity, it exits success
1. When ran without proper credentials, it exits success

This should be addressed and have proper return codes.

sharms:
For more information on exit status and the importance of it: http://tldp.org/LDP/abs/html/exit-status.html

rogeruiz:
Updated with some comments on my recent example here. https://github.com/18F/cg-deploy-stratos/pull/2/comment#issuecomment-411811220

cloudtrail-check.sh
timothy-spencer: I agree.  I think that adding it to a pipeline is not going to be as valuable as people think, because there are subjective parts to the audit, and it's just as easy to run a script as it is to go look at some pipeline output.

That said, it is not a bad thing to do.  Just extra work.  I'm happy to do that.

LinuxBozo: So no one else misses the point going forward: It's not valuable because the limit parameters are not valuable due to not being properly defined. There should be nothing subjective about it. Due to that, the value has absolutely nothing to do with automation or not.

timothy-spencer: Well, the whole point of this audit is that it is something that needs to be done by a human, right?  That's why it's in the tasks we must do when we are on support.  What this script does is automate the gathering of the information that we as operators are required to evaluate, not turn the audit into a "yes/no" script.  Doing that would take FAR more work:  https://xkcd.com/1425/

This is why I think that the utility of creating a pipeline and so on is marginal.  Again, not opposed to doing that.  Just trying to make sure that people were aware that this script may be a solution to a different problem than they are thinking of.  I was trying to solve the problem of "man, paging through all this json takes a lot of time and could be done better by a computer, especially when we have these convenient rules laid out in our documentation".

LinuxBozo: The point of any audit, not just this one, isn't about it must be done by a human. It's that it exists, it has parameters that make sense and that **they can be properly evaluated** in the first place, regardless of whether it's a human or not. This is a blatant hole in the parameters of this particular check. It has nothing to do really with this script, or your work on it, but rather the original context. As such, we shouldn't be replicating something that doesn't make sense, but rather change the parameters as a team, and then re-evaluate scripting it.

PR #76, opened 2018-07-19
Add network.admin scope for container networking

rogeruiz:
See docs here;
https://docs.cloudfoundry.org/devguide/deploy-apps/cf-networking.html#grant

See error message from trying this without the scopes;
```
$ cf add-network-policy hw-py --destination-app hw-go
Adding network policy to app hw-py in org sandbox-gsa / space roger.ruiz as roger.ruiz@gsa.gov...
provided scopes [cloud_controller.read password.write cloud_controller.write openid scim.write scim.read cloud_controller.admin uaa.user] do not include allowed scopes [network.admin network.write]
FAILED
```

### DIFF
--- a/make-cf-admin.sh
+++ b/make-cf-admin.sh
@@ -31,7 +31,7 @@ if ! hash uaac 2>/dev/null; then
   gem install cf-uaac
 fi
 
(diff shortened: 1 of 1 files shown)


rogeruiz:
Follow along in the Slack thread, where we discuss this further: https://gsa-tts.slack.com/archives/C0ENP71UG/p1532021345000167

PR #81, opened 2018-11-05
Add in some docs for how to operate count-sandbox-logs.py

rogeruiz:
n/t


rogeruiz:
I'll take care of the `snyk` issue we're having here.

PR #84, opened 2019-01-29
Add new scripts to suspend and activate better

rogeruiz:
This work is referenced in the following Runbook PR 18F/cg-site#1380

### DIFF
--- a/activate-org.sh
+++ b/activate-org.sh
@@ -5,7 +5,7 @@ set -e -x
 if [ "$#" -ne 1 ]; then
   echo
   echo "Usage:"
-  echo "   ./suspend-org.sh <org name>"
+  echo "   ./activate-org.sh <org name>"
   echo
     exit 1
 fi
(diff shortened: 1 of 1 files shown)


suspend-org.sh
```diff
@@ -10,11 +10,15 @@ if [ "$#" -ne 1 ]; then
...
-GUID=$(cf org $1 --guid)
+org_name=$1
+GUID=$(cf org "${org_name}" --guid)
 
-cf curl /v2/organizations/$GUID -X PUT -d '{"status":"suspended"}'
+cf curl "/v2/organizations/${GUID}" -X PUT -d '{"status":"suspended"}'
 
-for route in $(cf curl /v2/routes?q=organization_guid:$GUID | jq -r .resources[].metadata.guid)
-do
-  cf curl -X DELETE /v2/routes/$route?recursive=true
```
rogeruiz: We don't actually want to run any deletes on anything when we do this, because we always want to try to unsuspend things in the future. See :eyes: 18F/cg-site#1380

PR #90, opened 2019-04-17
[WIP] - Script for migrating to cflinuxfs3 for a specified org/space

jontours:
Might modify to run this migration for all orgs/spaces.  Thoughts?

### DIFF
--- a/cf-migrate-apps.sh
+++ b/cf-migrate-apps.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+if [ "$#" -lt 1 ]; then
+  printf "Usage:\n\n\t\$./cf-migrate-apps.sh <app_org/app_space>\n\n"
+  exit 1
+fi
+APP_PATH=$1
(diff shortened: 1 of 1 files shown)


cf-migrate-apps.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+if [ "$#" -lt 1 ]; then
+  printf "Usage:\n\n\t\$./cf-migrate-apps.sh <app_org/app_space>\n\n"
+  exit 1
+fi
+APP_PATH=$1
+for app in $(cf audit-stack | grep $APP_PATH | grep cflinuxfs2 | awk '{print $1}' | awk -F/ '{print $3}'); do cf change-stack $app cflinuxfs3; done
```
bengerman13: I think the grep for `$APP_PATH` should be anchored so we don't accidentally hit overlapping names.

cf-migrate-apps.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+if [ "$#" -lt 1 ]; then
```
bengerman13: how about `"$#" -ne 1` ?

bengerman13:
maybe I'm paranoid, but I'd like this better if it had a confirmation of what was going to be done, or an option to print what it would do instead of doing it.

PR #97, opened 2019-07-15
Add some audit scripts

bengerman13:
This adds two scripts that might be useful for future auditing:
- bosh-audit.py gets bosh events with optional time and user filters
- cf-audit.py gets cloudfoundry events with optional time and user filters


bosh-audit.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
+
```
hillaryj: Could we add an overview docstring with the description from the PR here?

hillaryj: A description is in the ArgParser description but would be nice to overview for future devs

cf-audit.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python3
+
```
hillaryj: Overview docstring here like in bosh-audit.py?

bosh-audit.py
bengerman13: that seems smart

mxplusb: Can you append that to the argparse help? Since you're at it...

bengerman13: Done

PR #105, opened 2019-12-06
Lowercase the totp username for reset

hillaryj:
The TOTP reset is case-sensitive, so  this makes it so the script forces input to lowercase.

### DIFF
--- a/reset-totp.sh
+++ b/reset-totp.sh
@@ -9,7 +9,8 @@ if [ "$#" -ne 2 ]; then
 fi
 
 deployment=$1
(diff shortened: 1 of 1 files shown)


bengerman13:
I know this was merged already, but I'm not confident it's the right move - we have some users with caps in their usernames.

hillaryj:
I looked into it further and yes, we do have users with capitalized usernames :( I'll revert the change and maybe instead have it leave a note that usernames are case sensitive.

hillaryj:
Resolved in #106

PR #110, opened 2020-01-16
add bucket size checker

bengerman13:
This adds a script to check bucket sizes. This is to help respond to a request from a customer.

This is a first-stab, and definitely has room for improvement. 
- I  think I'd like (later) to add functionality to create service keys, so it doesn't fail on an unbound service instance.
- machine-readable output would be nice

This also adds some functions that I think may be reusable elsewhere in cg-scripts, so I pulled them out into a source-able file

# Security Considerations
None.

### DIFF
--- a/get-bucket-sizes-for-org.sh
+++ b/get-bucket-sizes-for-org.sh
@@ -1,0 +1,95 @@
+#!/bin/bash
+
+CG_SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+source ${CG_SCRIPTS_DIR}/handy-functions.sh
+
+function usage {
+    echo
+    echo "Usage:"
+    echo "  $0 [-e] -o <org>"
+    echo
+    echo "  Options:"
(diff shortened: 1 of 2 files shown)


mxplusb:
@bengerman13 can you paste the output of this? I never got this working very well.

bengerman13:
(org, space, and service names have been changed)
```
$ USE_ENV_CREDENTIALS=yes ./get-bucket-sizes-for-org.sh an-org
Space: space-one
    a-bucket: Total Objects: 0 Total Size: 0 Bytes
    another-bucket: Total Objects: 0 Total Size: 0 Bytes
    additional-01: Total Objects: 0 Total Size: 0 Bytes
    additional-02: Total Objects: 0 Total Size: 0 Bytes
Space: space-two
    s3-bucket: Total Objects: 0 Total Size: 0 Bytes
Space: space-three
    s3-bucket: Total Objects: 0 Total Size: 0 Bytes
Space: space-four
    bucket: Total Objects: 0 Total Size: 0 Bytes
```

mxplusb:
Great, thanks!

PR #111, opened 2020-01-17
allows getting bucket names based on service ids

bengerman13:
We build bucket names programmatically based on service ids. This change lets the bucket size fetcher get the sizes programmatically, too

# Security Considerations
None

### DIFF
--- a/get-bucket-sizes-for-org.sh
+++ b/get-bucket-sizes-for-org.sh
@@ -10,7 +10,7 @@ function usage {
     echo
     echo "  Options:"
     echo "      -e     :     Use existing credentials for AWS, rather than service binding credentials"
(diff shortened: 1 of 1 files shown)


get-bucket-sizes-for-org.sh
```diff
@@ -10,6 +10,7 @@ function usage {
     echo
     echo "  Options:"
     echo "      -e     :     Use existing credentials for AWS, rather than service binding credentials"
+    echo "      -p     :     Bucket prefix. Requires -e, and allows getting sizes for all buckets even if they're not bound"
```
mxplusb: You should clarify what that means because I'm looking at it and I'm not sure what it means lol.

bengerman13: I tried to clarify it a little.

PR #112, opened 2020-01-29
Parse nessus vulnerabilities for sanity

pburkholder:
Requires:

`pip3 install nessus-file-reader`

Creates output like:

```
python3 parse-nessus-xml.py Production_Vulnerability_scan_xxxx.nessus
File name: /Users/peterburkholder/Downloads/Production_Vulnerability_scan_xxxx.nessus
File size: 67.6 MiB

33851, Risk: Low, Plugin Name: Network daemons not managed by the package system, https://www.tenable.com/plugins/nessus/33851
	foo-production
        bar-production
        baz-production
```

Solves https://github.com/18F/cg-private/issues/112

### DIFF
--- a/parse-nessus-xml.py
+++ b/parse-nessus-xml.py
@@ -15,8 +15,10 @@ nessus_scan_file = sys.argv[1]
 root = nfr.file.nessus_scan_file_root_element(nessus_scan_file)
 file_name = nfr.file.nessus_scan_file_name_with_path(nessus_scan_file)
 file_size = nfr.file.nessus_scan_file_size_human(nessus_scan_file)
+start_date = nfr.scan.scan_time_start(root)
 print(f'File name: {file_name}')
 print(f'File size: {file_size}')
+print(f'Scan start date: {start_date}')
 print('')
 
 
--- a/parse-nessus-xml.py
+++ b/parse-nessus-xml.py
@@ -35,7 +35,7 @@ for report_host in nfr.scan.report_hosts(root):
         hosts.append(report_host_name)
         vulnids[id] = hosts
 
-for key in vulnids:
+for key in sorted(vulnids):
     if key.find("Risk: None") == -1 :
         affected_hosts = vulnids[key]
         print(key)
--- a/parse-nessus-xml.py
+++ b/parse-nessus-xml.py
@@ -35,6 +35,10 @@ for report_host in nfr.scan.report_hosts(root):
 
 for key in vulnids:
     if key.find("Risk: None") == -1 :
+        affected_hosts = vulnids[key]
         print(key)
-        for site in vulnids[key]:
-            print('\t{}'.format(site))
+        if len(affected_hosts) > 4:
+            print('\t{} affected hosts found ...'.format(len(affected_hosts)))
+        else:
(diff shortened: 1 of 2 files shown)


pburkholder:
I've gone over the code at https://github.com/LimberDuck/nessus-file-reader/ and didn't see anything suspicious in terms of leaking information to outside sources.  But this probably could be rewritten to not require any PIP install now that I've seen how it's done.

parse-nessus-xml.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+vulnids = {}
+for report_host in nfr.scan.report_hosts(root):
+    report_host_name = nfr.host.report_host_name(report_host)
+    for report_item in nfr.host.report_items(report_host):
+        plugin_id = int(nfr.plugin.report_item_value(report_item, 'pluginID'))
+        risk_factor = nfr.plugin.report_item_value(report_item, 'risk_factor')
+        plugin_name = nfr.plugin.report_item_value(report_item, 'pluginName')
+        id = '{}, Risk: {}, Plugin Name: {}, https://www.tenable.com/plugins/nessus/{}'.format(
```
bengerman13: I think an f-string would make this line a lot nicer

parse-nessus-xml.py
```diff
@@ -0,0 +1,40 @@
...
+                                            plugin_id, risk_factor, plugin_name, plugin_id)
+        hosts = vulnids.get(id, [])
+        hosts.append(report_host_name)
+        vulnids[id] = hosts
+
+for key in vulnids:
+    if key.find("Risk: None") == -1 :
+        print(key)
+        for site in vulnids[key]:
+            print('\t{}'.format(site))
```
bengerman13: either an f-string, or just `print('\t', site)`

bengerman13:
re: could do it on your own without the dependency - .nessus files are just XML, we can also export them as csvs if that's easier to work with

parse-nessus-xml.py
```diff
@@ -0,0 +1,44 @@
+import nessus_file_reader as nfr
+import os
+import glob
+import traceback
+import time
+import sys
+
+
+if len(sys.argv) == 1:
```
hillaryj: I generally recommend this goes inside a `if __name__ == "__main__":` block.

parse-nessus-xml.py
```diff
@@ -0,0 +1,44 @@
...
+
+for key in vulnids:
+    if key.find("Risk: None") == -1 :
+        affected_hosts = vulnids[key]
+        print(key)
+        if len(affected_hosts) > 4:
+            print('\t{} affected hosts found ...'.format(len(affected_hosts)))
+        else:
+            for site in affected_hosts:
+                print('\t{}'.format(site))
```
hillaryj: There should be an newline (aka empty line) at the end of the file per python style guides as a line/file delimiter - I've started running `black` on my files to have it take care of formatting per some discussion in the python or dev channels on Slack and it works really well.

pburkholder:
Thanks for the comments @hillaryj and @bengerman13. I'm merging as is for now, but I have more tricks I want to add so I'll act on your comments on the next iteration.

PR #116, opened 2020-02-28
[WIP] Useful scripts for user reset and pipeline updates

hillaryj:
- Adds a script for updating a pipeline config automatically with prompts of what to do next
  - The `KEY` isn't always correctly found, so the `awk` command should get refined
- Adds a script to strip user roles
  - I had grand plans to use the `cf curl` api to get all the user's org and space roles to iterate through them, but this version just takes what's specified on the command line

## security considerations
None, these incorporate the procedures already on cloud.gov that operators follow.

### DIFF
--- a/strip-user-org-and-space-roles.sh
+++ b/strip-user-org-and-space-roles.sh
@@ -1,22 +1,42 @@
-#!/bin/bash -ex
+#!/usr/bin/env bash
 
-# User input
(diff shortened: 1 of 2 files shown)


tammersaleh:
Thank you @hillaryj - I love the fly script.  I'll take the PR from here.

PR #119, opened 2020-03-26
Add s3-history script

tammersaleh:

Add script to show s3 object version history diffs

## security considerations

none


bengerman13:
This is cool!
I think I'd want to be able to pass date or version id ranges if I was going to use this a ton, but a working script is better than a wishlist :)

PR #127, opened 2020-04-21
add auditor creation script

bengerman13:
- add a script to make a cf auditor

## security considerations

Update: This improves security by better applying least privilege to appropriate roles. A CF admin user is a member of the following groups.

`"cloud_controller.admin" "uaa.admin" "scim.read" "scim.write" "admin_ui.admin" "network.admin"`

For read-only purposes this role only needs:

`"cloud_controller.global_auditor" "scim.read" "admin_ui.user"`


pburkholder:
Updated Security Considerations to:

---
This improves security by better applying least privilege to appropriate roles. A CF admin user is a member of the following groups.

"cloud_controller.admin" "uaa.admin" "scim.read" "scim.write" "admin_ui.admin" "network.admin"

For read-only purposes this role only needs:

"cloud_controller.global_auditor" "scim.read" "admin_ui.user"

spgreenberg:
Is the intention to apply this to an existing CF user, meaning they have other necessary scopes already? Or is the intention to make this user a global auditor with all necessary CF scopes? Without stripping existing scopes and adding only what we want, the behavior is unclear to me.

Depending on how we define this auditor, we might need to add: uaa.user, scim.userids, openid, idps.read, clients.read, routing.routes.read, routing.router_groups.read, and doppler.firehose.

https://docs.cloudfoundry.org/concepts/architecture/uaa.html#scopes

bengerman13:
My thinking was to add to an existing CF user, as we do with the make-cf-admin script. Sounds like I need to do more research and thinking on this, based on your feedback @spgreenberg

pburkholder:
Instead of researching the impact of the different scopes, would it suffice to roll this out and then see if we need other PRs based on what I find I can't do?

spgreenberg:
Works for me. Can we name it “add-global-cf-auditor.sh” to be more clear of
the intent?

PR #133, opened 2020-06-29
remove user guid from all groups when orphaned

spgreenberg:

Remove an orphaned user group from all uaa groups. This occurs when a user is added to groups via `uaac member add` then deleted via `cf delete-user`.

## security considerations

Cleans up orphaned users.

### DIFF
--- a/uaa/remove-user-guid-from-all-groups.sh
+++ b/uaa/remove-user-guid-from-all-groups.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+usage() {
+  echo "Usage: "
+  echo "  $0 <user-guid>"
+  echo " "
+  echo "  <user-guid>: The guid of the user to remove from all groups."
+  echo " "
+}
+
+main() {
+  if [[ ! $# -eq 1 ]]; then
+    usage
+    exit 1
+  fi
+  local user_guid=$1
+  local group_guids=$(uaac curl /Groups | awk '/BODY/{y=1;next}y' | jq -r '.resources[] | select(.members[] | select(.value | contains("'"${user_guid}"'"))) | .id')
+  for group_guid in $group_guids; do
+    echo "Removing User: ${user_guid} from group: ${group_guid}"
+    uaac curl -X DELETE /Groups/${group_guid}/members/${user_guid}
+  done
+}
+
+main "$@"


uaa/remove-user-guid-from-all-groups.sh
```diff
@@ -0,0 +1,24 @@
...
+  echo " "
+}
+
+main() {
+  if [[ ! $# -eq 1 ]]; then
+    usage
+    exit 1
+  fi
+  local user_guid=$1
+  local group_guids=$(uaac curl /Groups | awk '/BODY/{y=1;next}y' | jq -r '.resources[] | select(.members[] | select(.value | contains("'"${user_guid}"'"))) | .id')
```
bengerman13, line 17: can we check if the value _is_ `user_guid`, rather than if it contains `user_guid`?

spgreenberg: not based on the structure. `members` is an array with multiple objects that have the `value` property.

pburkholder:
Although we don't have a consistent bash coding standard, I'm inclined to generally include `set -euo pipefail`

uaa/remove-user-guid-from-all-groups.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+usage() {
+  echo "Usage: "
+  echo "  $0 <user-guid>"
+  echo " "
+  echo "  <user-guid>: The guid of the user to remove from all groups."
```
pburkholder, line 7: `The GUID of the user to remove from all groups (up to 100, pagination not yet built-in)`

pburkholder:
I've picked some nits, but LGTM.

uaa/remove-user-guid-from-all-groups.sh
spgreenberg: Not sure what you are commenting on.

pburkholder (dismissed):

uaa/remove-user-guid-from-all-groups.sh
pburkholder: Actually this is more of an issue than I initially thought. 

I was commenting on the usage text since, it would only remove the user from the first 100 groups returned by the `uaac curl` call. I don't know if we're close to that limit or not. 

It may be worth adding pagination if we're over, say, 50 groups total and may hit 100 in the next few years.  

(I initially thought it was only an issue if the user themself was in more than 100 groups, but I see that I was wrong there)

markdboyd:
This PR is over 4 years old. Closing for now since we've lived without it for that long, but we can re-open if people desire.

PR #134, opened 2020-07-28
Add plugin output, make executable, and view hosts <= 6

pburkholder:

- Add a shebang line to make executable
- Show more hostnames for Nessus parse
- Include plugin evidence for Nessus parse

## security considerations

Changes parsing  / summarization of local Nessus XML files. None

### DIFF
--- a/parse-nessus-xml.py
+++ b/parse-nessus-xml.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
+
 import nessus_file_reader as nfr
 import os
 import glob
@@ -33,6 +35,7 @@ for report_host in nfr.scan.report_hosts(root):
         plugin_id = int(nfr.plugin.report_item_value(report_item, 'pluginID'))
         risk_factor = nfr.plugin.report_item_value(report_item, 'risk_factor')
         plugin_name = nfr.plugin.report_item_value(report_item, 'pluginName')
+        plugin_output = nfr.plugin.report_item_value(report_item, 'plugin_output')
 
         description = f"{plugin_id}, Risk: {risk_factor}, Plugin Name: {plugin_name}, https://www.tenable.com/plugins/nessus/{plugin_id}"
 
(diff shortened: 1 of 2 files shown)


parse-nessus-xml.py
```diff
@@ -53,7 +57,8 @@
     if vuln_report[key]["risk_factor"] != "None":
         affected_hosts = vuln_report[key]["hosts"]
         print(vuln_report[key]["full_description"])
-        if len(affected_hosts) > 4:
+#        print(vuln_report[key]["plugin_output"])
```
pburkholder, line 60: This should be an option, but for now it's easier for me to uncomment this when I need verbose output.

parse-nessus-xml.py
```diff
@@ -53,7 +57,8 @@
...
+        if len(affected_hosts) > 6:
```
ccostino, line 61: What's the limit on the number of affected hosts found for?

parse-nessus-xml.py
ccostino: I'm assuming you'd have to wire up reading input from the command line to do that, which isn't difficult but would be a bit of extra effort; I've used the [`click` library](https://click.palletsprojects.com/) in the past in case you're interested!

ccostino (approved):
Thanks, @pburkholder!  Approving, but had a question about the affected host count - I see it switched from `4` to `6` so that's fine, but I'm curious why the limit is there in the first place.  What's the significance with that number vs. show all hosts?

parse-nessus-xml.py
pburkholder: It makes the output saner, for example:
```
Hypothetical vuln with > more 6 affected hosts, https://www.tenable.com/plugins/nessus/9999
	48 affected hosts found ...
Hypothetical vuln with less than 6 affected hosts, https://www.tenable.com/plugins/nessus/9999
	shibboleth-production-idp-0-cf-production
	shibboleth-production-idp-1-cf-production
	shibboleth-production-idp-2-cf-production
```

PR #137, opened 2020-08-13
Changes to cloudtrail script for ec and es audit

ChrisMcGowan:

- Added in four new categories for auditing EC and ES events -   `CreateReplicationGroup`, `ModifyReplicationGroup`, `CreateElasticsearchDomain`, and `UpdateElasticsearchDomainConfig`
- Added jq logic to filter out aws-broker dev events `startswith("52.222"`

## security considerations:

- We do filter out events from the aws-browker in dev now hence the IP filter add


pburkholder:
For the records, these are the AU-6(3) changes to `cloudtrail-check.sh` for our SCR.

PR #138, opened 2020-09-17
add tooling env login for uaa

spgreenberg:

Logs into opslogin (uaa) when in the tooling environment

## security considerations

None. We could do this manually via credhub.


ccostino (approved):
Thanks, @spgreenberg!

PR #141, opened 2020-10-13
output org, quota and usage info

spgreenberg:

Outputs org and query usage in CSV format for the BU team. For example: https://docs.google.com/spreadsheets/d/1yM8-oNL76_Jtx7BLUi4SPdJcg8mAWwKNBv-GqNnShWA/edit?usp=sharing

## security considerations

This outputs org names, quotas, and GUIDs. Treat it accordingly.

### DIFF
--- a/audit/get-sandbox-user-emails.sh
+++ b/audit/get-sandbox-user-emails.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# Get a list of all orgs names prepended with the word sandbox"
+orgs=($(cf orgs | grep '^sandbox'))
+
+echo "Username, Org GUID"
+
+# Loop through the array and look up quota details.
+for org in "${orgs[@]}"
+do
+    guid=$(cf org $org --guid)
+
+    emails=($(cf curl "/v2/organizations/$guid/users" | jq '.resources | .[].entity | .username'))
+    for email in "${emails[@]}"
+    do
+        echo "${email}, ${guid}"
+    done
+    sleep 1
+done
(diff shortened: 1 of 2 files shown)


mheadd (approved):
Nice! This will be very useful.

ccostino (approved):
Thank you, @spgreenberg!  Nice and straightforward, and easy to pipe or copy the output into something else to parse the data as needed.

audit/get-sandbox-user-emails.sh
```diff
@@ -0,0 +1,19 @@
...
+orgs=($(cf orgs | grep '^sandbox'))
+
+echo "Username, Org GUID"
+
+# Loop through the array and look up quota details.
+for org in "${orgs[@]}"
+do
+    guid=$(cf org $org --guid)
+
+    emails=($(cf curl "/v2/organizations/$guid/users" | jq '.resources | .[].entity | .username'))
```
ccostino, line 13: We might want to shift this to the `v3` API call - that may change the response output though, so double check that it's what you expect!

ccostino (commented):
Thanks for the addition, @mheadd!  We may want to swap to the `v3` API, though; the response is likely different from the `v2` response so the `jq` processing may need some tweaking.

audit/get-sandbox-user-emails.sh
mheadd: That's fair. I'll merge this, but will create an issue to update the script for the newer version of the CF API.

ccostino: Sounds good, thanks @mheadd!

PR #143, opened 2020-12-02
Update the activate and suspend org scripts

ccostino:
This changeset updates the active and suspend organization utility scripts so that they function correctly and utilize version 7 of the `cf` CLI as well as the V3 API.

## Changes proposed in this pull request
- Adds a check to make sure version 7 of the `cf` CLI is being used
- Switches to using the V3 API to manage the organizations and applications
- Provides some additional guidance and comments

## Security considerations
- None: this change only modifies (and fixes) the function of existing public utility scripts

### DIFF
--- a/activate-org.sh
+++ b/activate-org.sh
@@ -12,6 +12,7 @@ if [ "$#" -ne 1 ]; then
   echo "Usage:"
   echo "   ./activate-org.sh <org name>"
   echo
+
   exit 1
 fi
 
@@ -25,6 +26,7 @@ if [ "$cf_version" -lt "$min_cf_version" ]; then
   echo "You must update your cf CLI to at least version 7."
   echo "Please run: brew update && brew upgrade cf-cli@7"
   echo
+
   exit 1
 fi
 
--- a/activate-org.sh
+++ b/activate-org.sh
@@ -31,22 +31,15 @@ if [ "$cf_version" -lt "$min_cf_version" ]; then
 fi
 
 org_name=$1
-GUID=$(cf org "${org_name}" --guid)
+org_guid=$(cf org "${org_name}" --guid)
 
 # Set the organization to not be suspended.
-cf curl "/v3/organizations/${GUID}" -X PATCH -d '{"suspended": false}'
+cf curl "/v3/organizations/${org_guid}" -X PATCH -d '{"suspended": false}'
 
-cf target -o "${org_name}"
-
-# Space names can contain spaces, so a for loop won't work as it'll break to
-# the next line upon the first whitespace character it reaches, not just a
-# newline character.  A while loop processes a whole line.
-cf spaces | tail -n +4 | while read -r space; do
-  cf t -s "${space}" > /dev/null
-
-  # The output of running `cf apps` includes more information than just the
-  # application name, which is the only thing we need.
-  for application in $(cf apps | tail -n +4 | awk '{ print $1 }'); do
-    cf start "${application}"
-  done
-done
+# Display a note reminding the platform operator to inform the business unit or
+# customer directly to restart their own apps, as we don't keep track of what
+# was running prior to suspending an organization.
+echo
+echo "Organization ${org_name} has been reactivated."
+echo "Please notify the customer that they must restart their applications."
+echo
--- a/activate-org.sh
+++ b/activate-org.sh
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+# Activates an organization by setting its suspended status to false and starts
+# all stopped applications in all spaces of the organization.
+
+# Requires cf CLI version 7+ to work.
+
 set -e -x
 
 if [ "$#" -ne 1 ]; then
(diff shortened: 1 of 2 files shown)


activate-org.sh
```diff
@@ -1,24 +1,52 @@
 #!/bin/sh
 
+# Activates an organization by setting its suspended status to false and starts
+# all stopped applications in all spaces of the organization.
+
+# Requires cf CLI version 7+ to work.
```
apburnes, line 6: If the org developers are using cf cli v6, will us launching the apps in v7 affect their ability to manage the apps?

ccostino: No, it shouldn't; the requirement here is because of the switch to the V3 APIs and I wanted to make sure we're using compatible tools on our side when running things.

apburnes: 👍

activate-org.sh
```diff
@@ -1,24 +1,52 @@
...
-cf curl "/v2/organizations/${GUID}" -X PUT -d '{"status":"active"}'
+# Set the organization to not be suspended.
+cf curl "/v3/organizations/${GUID}" -X PATCH -d '{"suspended": false}'
 
 cf target -o "${org_name}"
-for space in $(cf spaces "${org_name}" | tail -n +4); do
+
+# Space names can contain spaces, so a for loop won't work as it'll break to
+# the next line upon the first whitespace character it reaches, not just a
+# newline character.  A while loop processes a whole line.
```
apburnes, line 43: TIL

activate-org.sh
```diff
@@ -1,24 +1,52 @@
...
+cf spaces | tail -n +4 | while read -r space; do
   cf t -s "${space}" > /dev/null
-  for application in $(cf apps | tail -n +5); do
+
+  # The output of running `cf apps` includes more information than just the
+  # application name, which is the only thing we need.
+  for application in $(cf apps | tail -n +4 | awk '{ print $1 }'); do
```
apburnes, line 49: I assume app names cannot have spaces.

spgreenberg, line 49: I do not think we should be starting apps. We don't know if the apps were running when suspended and do not track this information. This could lead to apps available that are not intended to be running.

apburnes (approved):
💯

activate-org.sh
ccostino: Ooooh good point.  In this case we should just flip the org to not being suspended then?

ccostino: I assume not as well, but I assumed that with the space names and then we ran into one with a space!  So perhaps not safe to assume?  @spgreenberg do you know?  I don't see anything in the documentation mentioning something about this.

spgreenberg: I think so. I would put it on the customer to start up apps. It is possible they have a start order as well.

ccostino: I'll update this then, thanks!

suspend-org.sh
```diff
@@ -1,24 +1,52 @@
...
+# Space names can contain spaces, so a for loop won't work as it'll break to
+# the next line upon the first whitespace character it reaches, not just a
+# newline character.  A while loop processes a whole line.
+cf spaces | tail -n +4 | while read -r space; do
   cf t -s "${space}" > /dev/null
-  for application in $(cf apps | tail -n +5); do
+
+  # The output of running `cf apps` includes more information than just the
+  # application name, which is the only thing we need.
+  for application in $(cf apps | tail -n +4 | awk '{ print $1 }'); do
```
spgreenberg, line 49: Same as the comment for line 44. Getting app names this way is brittle. It is better to use the API.

activate-org.sh
spgreenberg: This is brittle as output to the cli format will cause this to break. App names in a space should be fetched via the API.

suspend-org.sh
```diff
@@ -1,24 +1,52 @@
...
+# Set the organization to suspended.
+cf curl "/v3/organizations/${GUID}" -X PATCH -d '{"suspended": true}'
 
 cf target -o "${org_name}"
-for space in $(cf spaces "${org_name}" | tail -n +4); do
+
+# Space names can contain spaces, so a for loop won't work as it'll break to
+# the next line upon the first whitespace character it reaches, not just a
+# newline character.  A while loop processes a whole line.
+cf spaces | tail -n +4 | while read -r space; do
```
spgreenberg, line 44: Use the API to get the list of spaces, not the CLI commands

activate-org.sh
```diff
@@ -1,24 +1,52 @@
...
+# Set the organization to not be suspended.
+cf curl "/v3/organizations/${GUID}" -X PATCH -d '{"suspended": false}'
 
 cf target -o "${org_name}"
-for space in $(cf spaces "${org_name}" | tail -n +4); do
+
+# Space names can contain spaces, so a for loop won't work as it'll break to
+# the next line upon the first whitespace character it reaches, not just a
+# newline character.  A while loop processes a whole line.
+cf spaces | tail -n +4 | while read -r space; do
```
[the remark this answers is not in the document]
ccostino: Good point - what was there previously had broken because the output had changed; if anything, all of the requests should flip to just using the API.

suspend-org.sh
ccostino: Swapped, thank you for the suggestion!

suspend-org.sh
ccostino: Swapped this out for a much better way of doing this.

activate-org.sh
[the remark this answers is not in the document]
ccostino: Accounted for!

activate-org.sh
ccostino: Swapped out with API calls!

ccostino:
@apburnes and @spgreenberg, would you mind taking another quick look with the last changes I made?

- Simplified the scripts
- Swapped cf CLI usage for API calls
- Only use object GUIDs instead of names

Thanks!

apburnes (approved):
💯

PR #144, opened 2020-12-03
Add script to check memory quota and usage for a single org

mheadd:
- Add script to check the org memory quota and current usage for a single org.
- Useful when fielding requests to increase or decrease memory quotas for an org.

## security considerations
None

### DIFF
--- a/audit/get-org-mem-usage.sh
+++ b/audit/get-org-mem-usage.sh
@@ -28,10 +28,10 @@ fi
 org=$(cf curl v3/organizations/?names=$1)
 
 # Get the guid for the organization
-guid=$(echo $org | jq '.resources[] | .guid' | sed 's/\"//g')
+guid=$(echo $org | jq -j '.resources[] | .guid')
 
 # Get the quota for the organization.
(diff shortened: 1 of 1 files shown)


mheadd:
I use this script a lot when we get requests to increase or decrease memory quotas. Way faster than logging into Stratos. Thought it might be helpful for others.

audit/get-org-mem-usage.sh
```diff
@@ -0,0 +1,48 @@
...
+  then
+    echo "You must supply an org name"
+    exit 1
+fi
+
+# Get organization details
+org=$(cf curl v3/organizations/?names=$1)
+
+# Get the guid for the organization
+guid=$(echo $org | jq '.resources[] | .guid' | sed 's/\"//g')
```
apburnes, line 31: You can use the `-j` flag with jq to remove the quotes

apburnes (approved):
Looks good! Added a comment about using a flag so you don't have to clean up the quotes with `sed 's/\"//g'`

audit/get-org-mem-usage.sh
mheadd: Good call, @apburnes. Change made.

PR #145, opened 2020-12-16
Script to empty an S3 bucket

ccostino:
This changeset introduces a new script to help empty the contents of an S3 bucket so that the service can be deleted.

h/t @apburnes for the idea!

## Changes proposed in this pull request
- New shell script to help empty an S3 bucket

## Security considerations
This can be run by users of the platform, but it will only work if they have permissions in the org and space to create service keys; in other words, this script doesn't do anything extra that a normal user or operator could do on their own by typing the commands in manually.

### DIFF
--- a/empty-s3-bucket.sh
+++ b/empty-s3-bucket.sh
@@ -20,7 +20,7 @@ fi
 
 # Create a service key and retrieve the credentials from it to use with awscli.
 SERVICE_INSTANCE_NAME="$1"
-KEY_NAME="$SERVICE_INSTANCE_NAME-key"
+KEY_NAME="cg-operator-empty-s3-bucket-key"
 
 # If the service key already exists, the cf CLI will ignore it and carry on
 # without an error.
--- a/empty-s3-bucket.sh
+++ b/empty-s3-bucket.sh
@@ -21,6 +21,9 @@ fi
 # Create a service key and retrieve the credentials from it to use with awscli.
 SERVICE_INSTANCE_NAME="$1"
 KEY_NAME="$SERVICE_INSTANCE_NAME-key"
+
+# If the service key already exists, the cf CLI will ignore it and carry on
+# without an error.
 cf create-service-key $SERVICE_INSTANCE_NAME $KEY_NAME
 S3_CREDENTIALS=$(cf service-key $SERVICE_INSTANCE_NAME $KEY_NAME | tail -n +2)
 
@@ -31,3 +34,6 @@ export AWS_DEFAULT_REGION=$(echo $S3_CREDENTIALS | jq -r '.region')
 
 # Empty the bucket.
 aws s3 rm -r * s3://$BUCKET_NAME
+
+# Remove the service key.
+cf delete-service-key $SERVICE_INSTANCE_NAME $KEY_NAME -f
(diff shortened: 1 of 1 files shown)


apburnes (approved):

apburnes:
Looks good to me!!!

empty-s3-bucket.sh
```diff
@@ -0,0 +1,33 @@
...
+  echo "Usage:"
+  echo "   ./empty-s3-bucket.sh <service instance name>"
+  echo
+
+  exit 1
+fi
+
+# Create a service key and retrieve the credentials from it to use with awscli.
+SERVICE_INSTANCE_NAME="$1"
+KEY_NAME="$SERVICE_INSTANCE_NAME-key"
```
spgreenberg, line 23: The key name could conflict with an existing key. Not sure this matters.

empty-s3-bucket.sh
```diff
@@ -0,0 +1,33 @@
...
+cf create-service-key $SERVICE_INSTANCE_NAME $KEY_NAME
+S3_CREDENTIALS=$(cf service-key $SERVICE_INSTANCE_NAME $KEY_NAME | tail -n +2)
+
+export AWS_ACCESS_KEY_ID=$(echo $S3_CREDENTIALS | jq -r '.access_key_id')
+export AWS_SECRET_ACCESS_KEY=$(echo $S3_CREDENTIALS | jq -r '.secret_access_key')
+export BUCKET_NAME=$(echo $S3_CREDENTIALS | jq -r '.bucket')
+export AWS_DEFAULT_REGION=$(echo $S3_CREDENTIALS | jq -r '.region')
+
+# Empty the bucket.
+aws s3 rm -r * s3://$BUCKET_NAME
```
spgreenberg, line 33: I believe this should delete the service key created on line 24

empty-s3-bucket.sh
ccostino: Good thing to call out!  I tested this and the `cf` CLI just carries on with a message saying it's already there, but still returns an `OK` - it does not cause an issue.

empty-s3-bucket.sh
ccostino: Fair enough!  I will add that in, thanks @spgreenberg!

empty-s3-bucket.sh
ccostino: I added a comment to call this out directly.

empty-s3-bucket.sh
ccostino: Added this as a final step.

empty-s3-bucket.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# Remove the service key.
+cf delete-service-key $SERVICE_INSTANCE_NAME $KEY_NAME -f
```
spgreenberg, line 39: If you are going to ignore and use an existing key, line 25, you probably shouldn't actually delete it.

spgreenberg: You could give it some quite unique name like "cg-operators-bucket-cleaner" or similar. It would reduce the likelihood of a collision. Given keys are for an instance, you don't need to use the $SERVICE_INSTANCE_NAME if you don't want to.

ccostino: Addressed, thank you!

PR #146, opened 2020-12-16
Fix awscli s3 rm command

ccostino:
This changeset fixes the aws s3 rm command in the empty S3 bucket script.

- Fixed the AWS CLI command for S3
- Adds a note about buckets with a large amount of objects

## Security considerations
None; command fix


spgreenberg (approved):
looks good

PR #147, opened 2020-12-23
Add script to get all users in the system

ccostino:
This changeset adds a new script that can retrieve all of the current users in the system.  The script can output the information to STDOUT or a CSV file.

- Adds a new script to retrieve basic user information from the system

## Future considerations:
- This script could be expanded a bit more to pair users with orgs, which is what this is intended to serve

## Security considerations
- Script makes use of the public CAPI exposed in our system
- Script can create and store a CSV file locally containing usernames and GUID mappings for further information retrieval

### DIFF
--- a/get-all-users.sh
+++ b/get-all-users.sh
@@ -1,0 +1,35 @@
+#!/bin/sh
+
+# Retrieves all users in the system and outputs a CSV file.
+
+# Requires cf CLI version 7+ and jq to work.
+
+# NOTE:  Only works if there are 5,000 or less users in the system at this time.
+
+set -e -x
+
+# Check to see if the person has at least version 7 of the cf CLI; if not, let
+# them know to upgrade and exit.
+min_cf_version=7
+cf_version=$(cf version | awk '{print $3}' | cut -c 1)
+
+if [ "$cf_version" -lt "$min_cf_version" ]; then
+  echo
+  echo "You must update your cf CLI to at least version 7."
+  echo "Please run: brew update && brew upgrade cf-cli@7"
+  echo
+
+  exit 1
+fi
+
+# Retrieve all users as JSON.
+cf curl '/v3/users?per_page=5000' > all-users.json
+
+# Create a CSV file with a header row.
+echo '"username","presentation name","guid","created at","origin"' > all-users.csv
+
+# Append all of the user data into the CSV file.
+# This information is sorted first by username, then by origin, to help with
+# grouping.  Records without usernames are stripped out as they are not actual
+# users, they're system accounts and used internally.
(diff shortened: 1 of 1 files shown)


get-all-users.sh
```diff
@@ -0,0 +1,35 @@
...
+  echo
+  echo "You must update your cf CLI to at least version 7."
+  echo "Please run: brew update && brew upgrade cf-cli@7"
+  echo
+
+  exit 1
+fi
+
+# Retrieve all users as JSON.
+cf curl '/v3/users?per_page=5000' > all-users.json
```
mheadd, line 26: Should the script `rm` this file when done since it's only an intermediate artifact?

ccostino: It could and I thought about that; I kept it for the moment in case having the raw JSON around is still helpful for whatever reason, but that's also just a `cf curl` call away.

get-all-users.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# Create a CSV file with a header row.
+echo '"username","presentation name","guid","created at","origin"' > all-users.csv
+
+# Append all of the user data into the CSV file.
+# This information is sorted first by username, then by origin, to help with
+# grouping.  Records without usernames are stripped out as they are not actual
+# users, they're system accounts and used internally.
+cat all-users.json | jq -r '.resources |= sort_by(.username) | .resources |= sort_by(.origin) | .resources[] | select(.username != "") | [ .username, .presentation_name, .guid, .created_at, .origin ] | @csv' >> all-users.csv
```
mheadd, line 35: Wondering if it makes more sense to not redirect to a file within the script, as that would allow further processing on the output before saving it. For example: `./get-all-users.sh | grep epa.gov > all-users.csv`

ccostino: 🤔 maybe, that depends on what we ultimately want to do; I could also set the script to allow for both things. :-)

pburkholder (approved):
I'm going to approve this as tweaking it to make it better is probably not worth our effort at this point. I'd suggest you add a TODO comment secation at the top to later on:

- use pagination so there's not a 5000-user limit (and to use smaller pages)
- use mktemp to get a unique temporary filename for the json out
- use a trap to make sure the temporary file is removed.

We may find that if we use this we'll want to rewrite in python to support options, but it serves a need, let's roll with it.

get-all-users.sh
ccostino: I updated the script and added options for this. 😃

ccostino:
Thanks for the feedback, @mheadd and @pburkholder!  I updated the script to account for these things and simplified it a bit, too.

mheadd (approved):
👍

PR #149, opened 2021-01-13
print the service instance creation date

spgreenberg:

Prints the service instance creation date in the output for billing purposes.

## security considerations

none


spgreenberg:
This script had some formatting issues with various service instances. I am not sure if that is because of the date. For the purpose of billing elastic search and redis, I instead created a different script.

PR #152, opened 2021-02-01
Add files via upload

kwadwok15:
- New script to pull all sandbox users without agreement.
- Script filters orgs based on org name. (non sandbox & sandbox)
- script out csv file into working folder. 

## security considerations
- Script pulls user names, which contains emails addresses. Keep this in mind when handling the script output

### DIFF
--- a/non_agreement_sandboxusers.sh
+++ b/non_agreement_sandboxusers.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+
+# Retrieves all users who are not part of an orgs with an agreement (sandbox users) and outputs to a CSV file.
+
+# Requires cf CLI version 7+ and jq to work.
+
+# NOTE:  The script logic is based on the assumtion that all sandbox orgs orgs do not have agreements.  
+
+# NOTE:  Only works if there are 5,000 or less orgs in the system at this time.
+
+
+set -e
+
+# Printing headers for csv file 
+printf 'User_Name,Agency,Account_creation_date,Org_Name\n' > users_to_contact.csv
+
+# Getting all non "sandbox-" org guids and storing it to a text file for later use 
+ORG_IDS=$(cf curl '/v3/organizations?per_page=5000'| jq -r '.resources |= sort_by(.name) | .resources[] | select(.name | contains("sandbox") | not ) | .guid')
+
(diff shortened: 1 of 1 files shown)


non_agreement_sandboxusers.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+
+# Retrieves all users who are not part of an orgs with an agreement (sandbox users) and outputs to a CSV file.
+
+# Requires cf CLI version 7+ and jq to work.
+
+# NOTE:  The script logic is based on the assumtion that all sandbox orgs orgs do not have agreements.
```
ccostino, line 7: Looks like an extra `orgs` here :-)

```suggestion
# NOTE:  The script logic is based on the assumption that all sandbox orgs do not have agreements.  
```

non_agreement_sandboxusers.sh
```diff
@@ -0,0 +1,63 @@
...
+                printf '%s, %s\n' "${ORG_USER_NAME}" "${ORG_NAME}" >> users_to_contact.csv
+                echo "users_to_contact csv printed successfully"
+            fi 
+        fi
+    done
+done
+
+# Removing file with non sandbox orgs user guids 
+echo "script executed sucessfully, removing temp file "
+rm -f non_sandbox_org_users.txt
```
ccostino, line 63: This isn't something that needs to be done now, but just for future consideration as it's something that @pburkholder brought to my attention as well - there are a couple of additional tools in UNIX that we can make use of for handling temporary files:

- [`mktemp`](https://ostechnix.com/the-mktemp-command-tutorial-with-examples-for-beginners/)
- [`trap`](https://www.shellscript.sh/trap.html)

`mktemp` is useful because it throws the file in `/tmp`, set it with a random/unique name, and gives it appropriate permissions, and `trap` ensures the file gets removed even if the script fails or is canceled halfway through.

Again, not something that needs to be done now, but something I wanted to share as I wasn't fully aware of these things either!

ccostino (commented):
Thanks, @kwadwok15!  There was only a small thing in the comments, and I added a note for future scripts, but otherwise this looks good!

ccostino (approved):
Thanks again, @kwadwok15 and @JohnGMullen!

PR #153, opened 2021-02-04
pushing cf-audit v3 changes

kwadwok15:
- update cf-audit.py to use use cf v3 api call
- update pagination logic to account for v3 pagination update. 
-
-

## security considerations
The script prints output to screen, which might show up in terminal history. 
The script print user emails (PII) to screen.

### DIFF
--- a/cf-audit.py
+++ b/cf-audit.py
@@ -16,6 +16,7 @@ import json
 import logging
 import subprocess
 import urllib
+from urllib.parse import urlparse
 
 
 def main():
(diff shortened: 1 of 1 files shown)


cf-audit.py
```diff
@@ -24,39 +25,44 @@ def main():
...
 
 def get_args():
     logging.basicConfig(level=logging.INFO)
     parser = argparse.ArgumentParser(
         description=description
     )
-    parser.add_argument('--after', help="find events after this timestamp (timestamp should be ISO8601)")
-    parser.add_argument('--before', help="find events before this timestamp (timestamp should be ISO8601)")
+    parser.add_argument('--after', help="find events after this timestamp (timestamp UTC formart YYYY-MM-DDThh:mm:ssZ)")
+    parser.add_argument('--before', help="find events before this timestamp (timestamp UTC formart YYYY-MM-DDThh:mm:ssZ)")
```
ccostino, line 65: Thanks for being a bit more explicit on the format here, that helps!  Just a slight spelling correction:

```suggestion
    parser.add_argument('--after', help="find events after this timestamp (timestamp UTC format YYYY-MM-DDThh:mm:ssZ)")
    parser.add_argument('--before', help="find events before this timestamp (timestamp UTC format YYYY-MM-DDThh:mm:ssZ)")
```

ccostino (commented):
Thank you, @kwadwok15!  This looks good to me as far as the changes, just a small spelling correction.

@pburkholder, you may also want to take a look at this to make sure the script is still providing the information needed for an audit.  We noticed that with the [API change from V2 to V3](http://v3-apidocs.cloudfoundry.org/version/3.93.0/#audit-events-in-v3) the event types were split out, and while it looks like the `audit events` are what we need, we just wanted to be sure.  Thanks!

ccostino (approved):
Thanks, @kwadwok15!

PR #154, opened 2021-02-12
updated cf-create-org to use v3 api

kwadwok15:
- changed logic for deleting "admin" from new created org. 
- 
-

## security considerations
none

### DIFF
--- a/cf-create-org.sh
+++ b/cf-create-org.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-set -e
+#set -e
+set -x
 
 if [ "$#" -lt 5 ]; then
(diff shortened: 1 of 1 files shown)


kwadwok15:
Changed the logic for deleting "admin" user from newly created org. 
New logic: v3 api does not associate users to an org but rather user is associate to a role and role is associated to an org. To achieve the same behavior the script 
1. pull "admin" user guid 
2. pull org guid 
3. pull role guid filtered by role_type"organization_user", admin user guid, and org guid
4. then delete the role based on role guid. 

Next step. Test the script in CF stage environment.

ccostino (commented):
This looks good so far, @kwadwok15!  Only one thing caught my eye, which was escaping some of the double quotes (`"`) - it looks like this might not be needed though, looking at the previous line that is being replaced?

kwadwok15:
@ccostino, noted

ccostino (approved):
Looks good, thanks @kwadwok15!

kwadwok15:
thanks @ccostino  merging PR now.

PR #158, opened 2021-02-24
updating get-bucket-size-for-org script to v3

### LINKED ISSUE #156: Update get-bucket-sizes-for-org script to use V3 of the CF API
This [script](https://github.com/cloud-gov/cg-scripts/blob/master/get-bucket-sizes-for-org.sh) makes calls to V2 of the CF API and needs to be updated to use V3.

Issue spawned from #142.

kwadwok15:
- changed all references of v2 endpoint to v3 endpoint
- added another api call to account for the changes made with v3 
- changed the way data is pulled to account for changes made with v3
- - fixes #156 

## security considerations
The script pulls bucket name, access_key_id and secret_access_key for s3 should be run by an approved admin user


mheadd (approved):
Looks good. Thanks, @kwadwok15!

PR #161, opened 2021-03-11
adding add-remove-pages-users script

kwadwok15:
- adding script to add and remove users to and from the pages group in uaa
-
-

## security considerations
The above scopes are arbitrary strings used to denote Pages accounts in UAA. They have no impact on security except in the Pages app.


spgreenberg:
Not all users will be admins. So I think we need to be able to add a pages.user separate from a pages.admin. This could be a flag or a separate script.

kwadwok15:
I was not sure if that was the case. I will add another flag to account for that.

PR #163, opened 2021-03-29
Validate admins for the current BOSH

pburkholder:
-
- gets the correct secret for Tooling and PRODUCTION.
-

## security considerations

Improvement - makes it easier to check admins in our envs

### DIFF
--- a/validate-admins.sh
+++ b/validate-admins.sh
@@ -1,38 +1,57 @@
-#!/bin/bash
+#!/bin/bash 
 
-if [ "$#" -ne 2 ]; then
-  echo
-  echo "Usage:"
-  echo "   ./validate-admins.sh <uaa-target> <uaa-admin-client-secret>"
-  echo
-  echo "   EX:  ./validate-admins.sh login.fr.cloud.gov S3c4Et"
-  echo 
-  echo "   Obtain uaa-admin-client-secret by running:"
-  echo 
-  echo "   credhub get -n \"/bosh/cf-{environment-name}/uaa_admin_client_secret\" | grep value | sed -r 's/value: //g'"
-  echo
-    exit 1
-fi
+# set -e -u -o pipefail
 
-uaac target $1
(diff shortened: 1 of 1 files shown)


validate-admins.sh
```diff
@@ -1,38 +1,57 @@
...
+        uaac users -a username "id eq '${uuid}'" | grep username | awk '{print $2}'
+    done
+
+    echo -e "\n\nadmins for concourse\n"
+    for uuid in $(uaac groups -a members "displayName eq 'concourse.admin'" | grep value | awk '{print $2}'); do 
+        uaac users -a username "id eq '${uuid}'" | grep username | awk '{print $2}'
+    done
+}
+
+case "$BOSH_DIRECTOR_NAME" in
```
bengerman13, line 31: two things, each maybe more work than they're worth:
- we could lowercase the director name and use it in the command to get the secret
- we could fetch the uaa url from credhub

validate-admins.sh
```diff
@@ -1,38 +1,57 @@
...
+    secret=$(credhub get -n /toolingbosh/opsuaa/uaa_admin_client_secret | grep value | sed -r 's/value: //g')
+    main opsuaa.fr.cloud.gov $secret
+    ;;
+  *)
+    if [ "$#" -ne 2 ]; then
+    echo
+    echo "Usage:"
+    echo "   ./validate-admins.sh <uaa-target> <uaa-admin-client-secret>"
+    echo
+    echo "   EX:  ./validate-admins.sh login.fr.cloud.gov S3c4Et"
```
bengerman13, line 46: this example says login.fr.cloud.gov - should we be using login on line 34? or is the example wrong?

bengerman13 (approved):
left a few comments for consideration, but it's fine as-is :)

validate-admins.sh
pburkholder: I don't know how to do those, so leaving as is.

PR #164, opened 2021-04-27
Add a script to retrieve all systems running with domains

ccostino:
This is a convenience script for operators to run to retrieve all systems running in the cloud.gov platform that have a domain associated with them.  The purpose of this script is to make it easier to pull system metrics too get an estimate of what is currently running.

The script assumes you are logged into CF in the command line, and outputs information in CSV format to be put in a spreadsheet.

- Adds a new script to retrieve all domains in the system

## Security considerations:
- Must be run by a platform operator
- Prints out org, space, and service instance names to the screen
- All information is available via the CF API for those who have access to the system

### DIFF
--- a/get-all-systems-with-domains.py
+++ b/get-all-systems-with-domains.py
@@ -1,75 +1,198 @@
 #!/usr/bin/env python
 
-"""
-Script to provide a listing of all domains found within the system, paired with
-the org and space names they belong to.
+description = """
+Script that provides a listing of all domains found within the system, paired
+with the org and space names they belong to and a bit of other metadata.  The
+output is sorted in chronological order by created_at date.
 
-NOTE:  This script assumes you are logged into CF on the command line.
+NOTE:  This script assumes you are logged into CF on the command line as it
+makes use of calls with the cf CLI.
 """
 
-import csv
+import argparse
 import json
+import logging
 import subprocess
 
-from urllib import error, request
-
-
+# CF API URI endpoint configuration.
 CF_SERVICE_INSTANCES_API_URI = "/v3/service_instances?service_plan_names=cdn-route,custom-domain,domain,domain-with-cdn&order_by=created_at"
 CF_SPACES_API_URI = "/v3/spaces/"
 CF_ORGANIZATIONS_API_URI = "/v3/organizations/"
+CF_SERVICE_PLANS_API_URI = "/v3/service_plans/"
+CF_SERVICE_OFFERINGS_API_URI = "/v3/service_offerings/"
 
 
-def get_all_instances():
+# Federalist organization name.
+FEDERALIST_ORG_NAME = "gsa-18f-federalist"
+
+
+# Logging configuration - this will output logs to the console/STDOUT/STDERR.
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+console_handler = logging.StreamHandler()
+console_handler.setLevel(logging.DEBUG)
+formatter = logging.Formatter("%(asctime)s - %(levelname)s: %(message)s")
+console_handler.setFormatter(formatter)
+logger.addHandler(console_handler)
+
+
+def main():
+    """
+    Runs the script.
+    """
+
+    args = get_args()
+
+    service_instances = get_all_instances(args.limit)
+    num_records = len(service_instances)
+
+    # Check to see if we should keep cloud.gov and Federalist domains together
+    # or make them separate and output the results.
+    if not args.split:
+        print("\n----- CSV OUTPUT ({0} TOTAL RECORDS) -----\n".format(
+            num_records
+        ))
+        output_instances(service_instances)
+    else:
+        print("\n----- CSV OUTPUT - SPLIT DOMAINS ({0} TOTAL RECORDS) -----".format(
+            num_records
+        ))
+        output_split_instances(service_instances)
+
+
+def get_args():
+    """
+    Configures and parses all of the arguments passed into the script.
+    """
+
+    parser = argparse.ArgumentParser(description=description)
+
+    parser.add_argument(
+        "--limit",
+        default=0,
+        help="Only process this amount of records",
+        type=int)
+    parser.add_argument(
+        "--split",
+        action="store_true",
+        default=False,
+        help="Split records between cloud.gov and Federalist domains"
+    )
+
+    return parser.parse_args()
+
+
+def get_all_instances(limit=0):
+    """
+    Retrieves all service instances that are associated with a domain configured
+    in the platform.
+    """
+
     service_instances = []
+    resources_parsed = 0
 
     page = CF_SERVICE_INSTANCES_API_URI
 
-    print("Beginning API processing...")
+    # If limit is set, display an extra message to remind the operator.
+    if limit > 0:
+        logger.info("Beginning API processing (limiting to {0} records)...".format(
+            limit
+        ))
+    else:
+        logger.info("Beginning API processing...")
 
     while page is not None:
         try:
             output = json.loads(subprocess.check_output(["cf", "curl", page]).decode("utf-8"))
-
-            for resource in output['resources']:
-                print("    ...Processing {0} ({1})...".format(resource["name"], resource["created_at"].split("T")[0]))
-
-                resource_info = {}
-                resource_info["service_plan_guid"] = resource["relationships"]["service_plan"]["data"]["guid"]
-                resource_info["domain_name"] = resource["name"]
-                resource_info["created_at"] = resource["created_at"].split("T")[0]
-
-                space_guid = resource["relationships"]["space"]["data"]["guid"]
-                space_output = json.loads(subprocess.check_output(["cf", "curl", CF_SPACES_API_URI + space_guid]).decode("utf-8"))
-
-                resource_info["space_name"] = space_output["name"]
-
-                org_guid = space_output["relationships"]["organization"]["data"]["guid"]
-                org_output = json.loads(subprocess.check_output(["cf", "curl", CF_ORGANIZATIONS_API_URI + org_guid]).decode("utf-8"))
-
-                resource_info["org_name"] = org_output["name"]
-
-                service_instances.append(resource_info)
-
-            page = output["pagination"].get("next")
-
-            if page is not None:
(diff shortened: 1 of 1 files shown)


get-all-systems-with-domains.py
```diff
@@ -0,0 +1,79 @@
...
+            for resource in output['resources']:
+                print("    ...Processing {0} ({1})...".format(resource["name"], resource["created_at"].split("T")[0]))
+
+                resource_info = {}
+                resource_info["service_plan_guid"] = resource["relationships"]["service_plan"]["data"]["guid"]
+                resource_info["domain_name"] = resource["name"]
+                resource_info["created_at"] = resource["created_at"].split("T")[0]
+
+                space_guid = resource["relationships"]["space"]["data"]["guid"]
+                space_output = json.loads(subprocess.check_output(["cf", "curl", CF_SPACES_API_URI + space_guid]).decode("utf-8"))
```
bengerman13, line 42: I think it would be good to split this onto multiple calls
Also, I think that `subprocess.check_output` and `json.loads` both have parameters to pass encodings, so we could drop `decode`

get-all-systems-with-domains.py
```diff
@@ -0,0 +1,79 @@
...
+
+            page = output["pagination"].get("next")
+
+            if page is not None:
+                uri_parts = page["href"].split("/")
+                page = "{0}/{1}".format(uri_parts[-2], uri_parts[-1])
+                print("...Processing next page...")
+            else:
+                print("...Finished processing.")
+        except subprocess.CalledProcessError as exc:
```
bengerman13, line 61: I think this except should be up right after the try.

Also, since this is a script to be run manually, and it's just running shell commands, I wonder if it's better to just let it crash on exception

get-all-systems-with-domains.py
```diff
@@ -0,0 +1,79 @@
...
+
+                resource_info["org_name"] = org_output["name"]
+
+                service_instances.append(resource_info)
+
+            page = output["pagination"].get("next")
+
+            if page is not None:
+                uri_parts = page["href"].split("/")
+                page = "{0}/{1}".format(uri_parts[-2], uri_parts[-1])
```
bengerman13, line 57: I'm pretty sure you can use the `href` with the protocol and host within cf curl

get-all-systems-with-domains.py
```diff
@@ -0,0 +1,79 @@
...
+    service_instances = get_all_instances()
+
+    for instance in service_instances:
+        output = "{0},{1},{2},{3},{4}".format(
+            instance["service_plan_guid"],
+            instance["org_name"],
+            instance["space_name"],
+            instance["domain_name"],
+            instance["created_at"]
+        )
```
bengerman13, line 77: we imported csv, but then manually built a csv - can we either drop the import or use it here?

get-all-systems-with-domains.py
```diff
@@ -0,0 +1,79 @@
...
+    page = CF_SERVICE_INSTANCES_API_URI
+
+    print("Beginning API processing...")
+
+    while page is not None:
+        try:
+            output = json.loads(subprocess.check_output(["cf", "curl", page]).decode("utf-8"))
+
+            for resource in output['resources']:
+                print("    ...Processing {0} ({1})...".format(resource["name"], resource["created_at"].split("T")[0]))
```
bengerman13, line 34: can we send this to stderr (either by pointing it there, or by using the `logging` library) to make this script more pipe and redirect friendly?

(same on 25)

get-all-systems-with-domains.py
```diff
@@ -0,0 +1,79 @@
...
+the org and space names they belong to.
+
+NOTE:  This script assumes you are logged into CF on the command line.
+"""
+
+import csv
+import json
+import subprocess
+
+from urllib import error, request
```
bengerman13, line 14: this import isn't used

bengerman13 (commented):
Thanks for taking the time to put this together!

I added some suggestions throughout. I think overall this might also be easier to read and reuse if we separated the fetching and parsing logic so it was more like

```
def get_all_instances():
    instances = []
    page = ...
    while page is not None:
        output = subproces ...
        instances.extend(output["resources"])
        page = ....
    return instances

def parse_instance(instance):
    # mess with instance
    return f" ..."

def main():
    for instance in get_all_instances():
        print(parse_instance(instance))

````

File: get-all-systems-with-domains.py (hunk 4 above)
Comment (ccostino) [reply get-all-systems-with-domains.py:77]: D'oh!  I thought I had cleaned up unused stuff at least. :-)

I had grander designs, if I had more time...

File: get-all-systems-with-domains.py (hunk 6 above)
Comment (ccostino) [reply get-all-systems-with-domains.py:14]: Really need a `d'oh` emoji here in GitHub...  Same as the `csv` import remark above!

File: get-all-systems-with-domains.py (hunk 3 above)
Comment (ccostino) [reply get-all-systems-with-domains.py:57]: Ah, but it didn't!  It might need to be formally wrapped in double quotes, though.

File: get-all-systems-with-domains.py (hunk 2 above)
Comment (ccostino) [reply get-all-systems-with-domains.py:61]: Definitely room for improvement here - referring back to what you had highlighted above about splitting things out, that'd help with more organized error handling where it makes sense.

**Comment by ccostino** [comment]
Thanks, @bengerman13!  It was definitely a put it together and make it work quick effort first and foremost. :-)

I'll take a look through to make some adjustments and incorporate your feedback, I could certainly use the change in activity!

Another thought I had was making it possible to split out cloud.gov-only and Federalist domains too, I think that's possible without too much more work.

File: get-all-systems-with-domains.py [hunk 7]
```diff
@@ -0,0 +1,236 @@
...
+
+    # Print out the two separate lists.
+    print("\ncloud.gov-only domains ({0} records):\n".format(
+        len(cloud_instances)
+    ))
+    output_instances(cloud_instances)
+    print("\nFederalist-only domains ({0} records):\n".format(
+        len(federalist_instances)
+    ))
+    output_instances(federalist_instances)
```
bengerman13, line 232: This makes it so this script doesn't output a CSV. To make it so this script outputs a CSV, what if we did one or both of these:
- include a column for `isfederalist`
- add a flag to exclude federalist and a flag for only federalist

get-all-systems-with-domains.py
```diff
@@ -0,0 +1,236 @@
...
+
+    args = get_args()
+
+    service_instances = get_all_instances(args.limit)
+    num_records = len(service_instances)
+
+    # Check to see if we should keep cloud.gov and Federalist domains together
+    # or make them separate and output the results.
+    if not args.split:
+        print("\n----- CSV OUTPUT ({0} TOTAL RECORDS) -----\n".format(
```
bengerman13, line 52: I think we should drop the printed metadata lines, or make them log statements instead. 
The reason is that by printing these, if I run `python get-all-systems-with-domains.py > domains.csv` then `domains.csv` is not a valid csv file.

A user can use `wc` if the want to know how many records there are, or open the csv with excel. If we need to signal to users that output will be in CSV format, we can do that in the help (although I think most folks who run this would probably immediately recognize csv output when they see it)

get-all-systems-with-domains.py
```diff
@@ -0,0 +1,236 @@
...
+    """
+
+    logger.info("    ...Parsing {0} ({1})...".format(
+        resource["name"],
+        resource["created_at"].split("T")[0]
+    ))
+
+    # Start pulling out the service instance information directly.
+    resource_info = {}
+    resource_info["domain_name"] = resource["name"]
```
bengerman13, line 161: is this actually the domain name, or is this the service instance name?

pburkholder:
Ben and Carlo -- Thanks for putting in the work to make this potentially reusable. --P

get-all-systems-with-domains.py
ccostino: 🤔 now that you point that out, technically speaking this is the service instance name name, but because of how the domain broker(s) work, the name seems to be a mirror of the domain (or the DNS record that is set for the domain).  I don't think the service instance metadata has the actual domain name in the output when I was looking and this was a close enough approximation.

Are you thinking it's better to be accurate in the description/key name of this, or somehow get the actual domain name?

get-all-systems-with-domains.py
ccostino: Okay that's definitely fair - be more UNIX-like!  Was getting too fancy with the output. :-)

get-all-systems-with-domains.py
ccostino: Good call - I'll think on this when making adjustments for the point below about the metadata lines.  Thanks!

get-all-systems-with-domains.py
bengerman13: yeah - I should've been clearer about that earlier when I mentioned print :)

get-all-systems-with-domains.py
bengerman13: the name is whatever the user wants it to be - it just happens that our biggest user (federalist) is very, very consistent about naming

I think just renaming the variable so it's clear that it's _not_ the domain name is what we need (and the comment above)


This also reminded me: printing a header row would be good, too

ccostino: Sounds good!  And yes to the header row - I'll have it include it by default but include an option to omit just in case it's helpful.

bengerman13: 👨‍🍳❤️👌

get-all-systems-with-domains.py
```diff
@@ -0,0 +1,283 @@
...
+
+def get_fiscal_year(date_string):
+    """
+    Figures out and returns the fiscal year label based on a provided calendar
+    date.
+    """
+
+    year, month, day = date_string.split("-")
+    new_fiscal_year_months = ["10", "11", "12"]
+    fiscal_year = int(year[2:])
```
bengerman13, line 253: well this is gonna blow up in 79 years /s

bengerman13 (approved):

get-all-systems-with-domains.py
ccostino: I was about to follow up with another sarcastic remark, but watch us eat our words and this still be used 79 years from now...

PR #165, opened 2021-05-05
Make my parsing scripts more useful.

pburkholder:

- Make sure the Nessus findings for unmanaged daemons are the ones we know to be generally bosh-managed
- Update Eddie to Ron
- Fix the Owasp script to not die if `wascid` omitted
- Skip non-cloud-gov sites in OWasp

## security considerations

Doesn't impact system operations. The content of the PR could provide clues to nefarious actors.  This is an older PR and the K8s related ones could be removed from the RegExes.

### DIFF
--- a/parse-nessus-xml.py
+++ b/parse-nessus-xml.py
@@ -28,6 +28,17 @@ start_date = nfr.scan.scan_time_start(root)
 print(f'File name: {file_name}')
 print(f'File size: {file_size}')
 print(f'Scan start date: {start_date}')
+DAEMONS = "bosh.dns|bosh.dns.adapter|bosh.dns.health|alertmanager|binding.cache|blackbox.exporter|bosh.exporter|cf.exporter"
+DAEMONS += "|concourse|docker|doomsday|firehose.exporter|gdn|gnatsd|grafana.server|java|kube.state.metrics|nessusd|nginx"
+DAEMONS += "|node.exporter|ntpd|oauth2.proxy|prometheus|pushgateway|ruby"
+DAEMONS += "|auctioneer|bbs|binding.cache|broker|cc.uploader|consul|discovery.registrar|dockerdt"
+DAEMONS += "|doppler|elasticsearch.exporter|etcd|file.server|flanneld|forwarder.agent|gorouter"
+DAEMONS += "|kube.apiserver|kube.controller.manager|kube.proxy|kube.scheduler|kubelet|locket"
+DAEMONS += "|log.cache|log.cache.cf.auth.proxy|log.cache.gateway|log.cache.nozzle|loggregator.agent|metrics.agent"
+DAEMONS += "|netmon|node|policy.server|policy.server.internal|prom.scraper|redis.server|rep"
+DAEMONS += "|reverse.log.proxy|rlp|reverse.log.proxy.gateway|rlp.gateway"
+DAEMONS += "|route.emitter|service.discovery.controller|silk.controller|silk.daemon"
+DAEMONS += "|ssh.proxy|statsd.injector|syslog.agent|tps.watcher|trafficcontroller|udp.forwarder|vxlan.policy.agent"
 
 vuln_report = {}
 for report_host in nfr.scan.report_hosts(root):
@@ -53,18 +64,6 @@ for report_host in nfr.scan.report_hosts(root):
             "plugin_output": plugin_output,
             "hosts": hosts
         }
-        DAEMONS = "bosh-dns|bosh-dns-adapter|bosh-dns-health|alertmanager|blackbox_exporter|bosh_exporter|cf_exporter"
-        DAEMONS += "|concourse|doomsday|firehose_exporter|gdn|gnatsd|grafana-server|java|kube-state-metrics|nessusd|nginx"
-        DAEMONS += "|node_exporter|ntpd|oauth2_proxy|prometheus|pushgateway|ruby"
-        DAEMONS += "|auctioneer|bbs|binding-cache|broker|cc-uploader|consul|discovery-registrar|dockerdt"
-        DAEMONS += "|doppler|elasticsearch_exporter|etcd|file-server|flanneld|forwarder-agent|goroutert"
-        DAEMONS += "|kube-apiserver|kube-controller-manager|kube-proxy|kube-scheduler|kubelet|locket"
-        DAEMONS += "|log-cache|log-cache-cf-auth-proxy|log-cache-gateway|log-cache-nozzle|loggregator-agent|metrics-agent"
-        DAEMONS += "|netmon|node|policy-server|policy-server-internal|prom-scraper|redis-server|rep"
-        DAEMONS += "|rlp|rlp-gateway|route-emitter|service-discovery-controller|silk-controller|silk-daemon"
-        DAEMONS += "|ssh-proxy|statsd-injector|syslog-agent|tps-watcher|trafficcontroller|udp-forwarder|vxlan-policy-agent"
-
-
 
         if plugin_id == 33851:
             for line in plugin_output.splitlines():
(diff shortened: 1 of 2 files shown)


parse-owasp-zap-xml.py
```diff
@@ -17,12 +17,14 @@
         for alert in site.findall('.//alertitem'):
             id = '{}, CWE id {}, WASC id {}, Risk {}, Plugin ID {}'.format(alert.find('name').text,
                                                     alert.find('cweid').text,
-                                                    alert.find('wascid').text,
+                                                    alert.findtext('wascid',"None"),
```
ccostino, line 20: Nice catch - should the rest of these also be switched or are you confident that they'll be there, or this should raise an error if they're not found?

parse-nessus-xml.py
```diff
@@ -28,10 +29,26 @@
...
+DAEMONS += "|concourse|docker|doomsday|firehose.exporter|gdn|gnatsd|grafana.server|java|kube.state.metrics|nessusd|nginx"
+DAEMONS += "|node.exporter|ntpd|oauth2.proxy|prometheus|pushgateway|ruby"
+DAEMONS += "|auctioneer|bbs|binding.cache|broker|cc.uploader|consul|discovery.registrar|dockerdt"
+DAEMONS += "|doppler|elasticsearch.exporter|etcd|file.server|flanneld|forwarder.agent|gorouter"
+DAEMONS += "|kube.apiserver|kube.controller.manager|kube.proxy|kube.scheduler|kubelet|locket"
+DAEMONS += "|log.cache|log.cache.cf.auth.proxy|log.cache.gateway|log.cache.nozzle|loggregator.agent|metrics.agent"
+DAEMONS += "|netmon|node|policy.server|policy.server.internal|prom.scraper|redis.server|rep"
+DAEMONS += "|reverse.log.proxy|rlp|reverse.log.proxy.gateway|rlp.gateway"
+DAEMONS += "|route.emitter|service.discovery.controller|silk.controller|silk.daemon"
+DAEMONS += "|ssh.proxy|statsd.injector|syslog.agent|tps.watcher|trafficcontroller|udp.forwarder|vxlan.policy.agent"
```
ccostino, line 42: Could this be a list instead, and you could do something like `DAEMONS.join('|') in the `re.search` calls below?  Just to help make this a bit more manageable when adding/removing things!

parse-nessus-xml.py
```diff
@@ -49,24 +66,46 @@
...
+                    daemon_count += 1
+                    continue
+                if re.search(rf'^/var/vcap/data/packages/({DAEMONS})/[0-9a-f]+/bin/{DAEMONS}$', line):
+                    daemon_count += 1
+                    continue
+                print("== Unknown daemon found: ",report_host_name,": ", line)
+
         vuln_report[plugin_id] = this_vuln
 
+print("Known deamons seen: ", daemon_count)
```
ccostino, line 90: Spelling error:

```suggestion
print("Known daemons seen: ", daemon_count)
```

parse-nessus-xml.py
```diff
@@ -49,24 +66,46 @@
...
 for key in sorted(vuln_report):
     if vuln_report[key]["risk_factor"] != "None":
         affected_hosts = vuln_report[key]["hosts"]
+        affected_hosts.sort()
         print(vuln_report[key]["full_description"])
 #        print(vuln_report[key]["plugin_output"])
         if len(affected_hosts) > 6:
             print('\t{} affected hosts found ...'.format(len(affected_hosts)))
         else:
+            #for site in affected_hosts.sort():
```
ccostino, line 102: Fair to remove this since it's being done above?

```suggestion
```

ccostino (commented):
Thanks, @pburkholder!  Overall it looks good to me, I just had a few quick questions.

parse-nessus-xml.py
```diff
@@ -28,10 +29,26 @@
 print(f'File size: {file_size}')
 print(f'Scan start date: {start_date}')
 
+DAEMONS = "bosh.dns|bosh.dns.adapter|bosh.dns.health|alertmanager|binding.cache|blackbox.exporter|bosh.exporter|cf.exporter"
+DAEMONS += "|concourse|docker|doomsday|firehose.exporter|gdn|gnatsd|grafana.server|java|kube.state.metrics|nessusd|nginx"
+DAEMONS += "|node.exporter|ntpd|oauth2.proxy|prometheus|pushgateway|ruby"
+DAEMONS += "|auctioneer|bbs|binding.cache|broker|cc.uploader|consul|discovery.registrar|dockerdt"
```
bengerman13, line 35: should `dockerdt` be `dockerd`?

parse-nessus-xml.py
```diff
@@ -49,24 +66,46 @@
...
+                    continue
+                if "The following running daemons are not managed by dpkg" in line:
+                    continue
+                if (line == "/bin/kube2iam"):
+                    daemon_count += 1
+                    continue
+                if re.search(rf'/var/vcap/bosh/bin/(bosh-agent|monit)', line):
+                    daemon_count += 1
+                    continue
+                if re.search(rf'^/var/vcap/data/packages/({DAEMONS})/[0-9a-f]+/bin/{DAEMONS}$', line):
```
bengerman13, line 83: maybe over-optimization for this script, but if you compile the regex outside the loop, this will run significantly faster

parse-nessus-xml.py
```diff
@@ -49,24 +66,46 @@
...
         }
+
+        # Match on known daemons:
+        if plugin_id == 33851:
+            for line in plugin_output.splitlines():
+                if (len(line) < 1):
+                    continue
+                if "The following running daemons are not managed by dpkg" in line:
+                    continue
+                if (line == "/bin/kube2iam"):
```
bengerman13, line 77: is this still a known daemon?

parse-nessus-xml.py
bengerman13: :pixar-ooo:

pburkholder: Yes. That would make maintenance a lot easier too. I can see already that I can remove a lot of k8s-related daemons.

parse-nessus-xml.py
pburkholder: It runs fast enough, so I'm going to keep the RE right where it's used.

parse-owasp-zap-xml.py
pburkholder: Next time it happens I'll change all of them. I hope to actually retire this with a real tool like DefectDojo, so not wanting to over-invest there.

pburkholder:
@bengerman13 @ccostino I accepted your suggestions,  rolled forward to Ashley (for now), and removed kube* stuff. OK to merge?

parse-nessus-xml.py
```diff
@@ -28,10 +29,88 @@
...
+    bosh.exporter
+    cf.exporter
+    concourse
+    docker
+    doomsday
+    firehose.exporter
+    gdn
+    gnatsd
+    grafana.server
+    java
```
bengerman13, line 48: `java` (and `ruby`) could be run as daemons for unknown processes, right? Do they belong here?

I think we should probably remove or refine them, depending on what the report shows. Either the report shows just the process without any params, in which case we don't know whether this is `java org.elasticsearch.Logstash` or `com.nefarious.Badstuff`. Or the report shows the process with params, in which case the anchored regex means we'll only ever match if we're running java as a daemon by itself

Maybe I'm overthinking

pburkholder: Now you got me overthinking this, but I got to clean up some stuff and make this more rigorous in terms of name AND path.

Unfortunately, I python `re` doesn't support subpattern recursion or I could do something Perl-ish like:

```
\/var\/vcap\/data\/packates\/({DAEMONS})\/[a-z0-9]+\/(?1)
```
and ensure the binary matched the name of the directory, but that doesn't seem easy at this point (and not worth the effort)

bengerman13: @pburkholder I have good or bad news, depending on your perspective.

Python's `re` does support backreferences, including named backreferences

bengerman13 (approved):

bengerman13:
LGTM

PR #170, opened 2021-08-04
add script for cdn

bengerman13:
- add script for cdn cert rotations
- fix wait_print newlines

Probably worth noting: this script does not attempt to delete the old IAM certificate. This is because we'd have to wait for Cloudfront to process the update, and I didn't want the script hanging for 30 minutes. 

We should just run through periodically and nuke expired IAM certs in commercial, anyway

## security considerations
None

### DIFF
--- a/certbot/README.md
+++ b/certbot/README.md
@@ -2,7 +2,18 @@
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


ccostino (approved):
Looks good to me, thanks @bengerman13!

PR #171, opened 2021-08-06
add script for all cdn renewals

bengerman13:
- Add a script to attempt to renew all expiring cdn certs
- Fix cleanup script for cases where the filename begins with `-`
- Fix multiple domain parsing for ALBs
- Add script to clean up expired certificates from iam

## security considerations
Automating means we're less likely to make mistakes from typos, etc, but also amplifies our ability to make bigger errors

### DIFF
--- a/certbot/renew_all_cdns.sh
+++ b/certbot/renew_all_cdns.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+## Calculate the days to expiration from now
+function days_to_cert_expiration {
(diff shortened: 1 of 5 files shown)


certbot/renew_all_cdns.sh
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
```
bengerman13, line 1: this script is pulled almost wholesale from our prometheus pipeline

ccostino (approved):
Thanks, @bengerman13!

PR #172, opened 2021-08-09
add script to renew all albs at once

bengerman13:
- add script to renew all albs at the same time
- add docs about the new scripts
- update alb_certs.py to bail instead of fail

## security considerations
None

### DIFF
--- a/certbot/README.md
+++ b/certbot/README.md
@@ -1,8 +1,32 @@
 # Less-manual manual cert rotation tools
 
-## Usage
+## all-in-one
 
-Both scripts expect you have set the environment variable CERTBOT_BUCKET_NAME to be the name
+*Note* This uses bash-isms not available on the default OS X bash (3.x). 
+In order to run either script, make sure you've installed a newer version of
+bash with homebrew
+
+In this directory, you'll find 2 scripts for renewing all the expiring certificates
+for albs and cdns: `renew_all_albs.sh` and `renew_all_cdns.sh`. These are based
+heavily on the scripts used by prometheus to detect expiring certs, so the scripts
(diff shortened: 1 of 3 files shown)


certbot/README.md
```diff
@@ -1,8 +1,32 @@
 # Less-manual manual cert rotation tools
 
-## Usage
+## all-in-one
 
-Both scripts expect you have set the environment variable CERTBOT_BUCKET_NAME to be the name
+*Note* This uses bash-isms not available on the default OS X bash (3.x).
```
ccostino, line 5: Missing words:

```suggestion
*Note* This uses bash-isms, which is not available on the default OS X bash (3.x). 
```

certbot/README.md
```diff
@@ -1,8 +1,32 @@
 # Less-manual manual cert rotation tools
 
-## Usage
+## all-in-one
 
-Both scripts expect you have set the environment variable CERTBOT_BUCKET_NAME to be the name
+*Note* This uses bash-isms not available on the default OS X bash (3.x). 
+In order to run either script, make sure you've installed a newer version of
+bash with homebrew
```
ccostino, line 7: Homebrew instructions for installing BASH:

```suggestion
bash with homebrew: `brew install bash`
```

ccostino (approved):
Thanks, @bengerman13!  Two minor suggestions if they're helpful, otherwise :shipit:

certbot/README.md
bengerman13: this suggestion doesn't work, but it does tell me that my original wording isn't great, so I'll tweak it

ccostino: No worries!

PR #187, opened 2021-12-02
Improve active sandbox and space count script

ccostino:
This changeset adds a few improvements to the active sandbox and space count script by allowing an operator to specify start and end dates to filter by.  If a start date (and an end date) are given, the results are constrained to those dates in an inclusive fashion: starting at 00:00:00 on the start date and ending at 23:59:59 on the end date (if provided).

Including the dates will also provide a new line of output, how many sandbox spaces were created during that time.

- Add support for accepting `--start-date` and `--end-date` arguments
- Add extra output when running the script with dates
- Add some help text when supplying a `-? / -h / --help` argument

## Security considerations:
- None; utilizes CAPI calls that all users of the system have access to

### DIFF
--- a/audit/get-active-sandbox-and-space-count.sh
+++ b/audit/get-active-sandbox-and-space-count.sh
@@ -2,44 +2,100 @@
 
 set -e
 
-# Counts the number of current active sandboxes in the platform.
+# Provides counts for the following:
+# - The total number of sandboxes spaces currently in the platform
+# - The number of current active sandboxes in the platform
+# - Optionally provides number of sandboxe spaces created with provided dates or
+#   a date range
 
 # An active sandbox is defined as any user's sandbox space with 1 or more
-# running apps in it.
+# running apps in it.  Sandboxes are cleared out every 90 days from the time of
+# first activity, so we can only ever count what is currently running.  That
+# said, we can determine these counts by using the cf CLI and CAPI to retrieve
+# all sandbox orgs, then filter a list of all apps with those corresponding org
+# GUIDs.
 
-# We can determine this by using the cf CLI and CAPI to retrieve all sandbox
-# orgs, then filter a list of all apps with those corresponding org GUIDs.
+# Sandbox spaces are defined as any space within a sandbox organization, e.g.,
+# all spaces under the sandbox-gsa org.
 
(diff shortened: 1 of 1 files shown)


audit/get-active-sandbox-and-space-count.sh
```diff
@@ -8,7 +10,38 @@
 # We can determine this by using the cf CLI and CAPI to retrieve all sandbox
 # orgs, then filter a list of all apps with those corresponding org GUIDs.
 
+if [[ $1 == "--help" ]]; then
+  echo " "
+  echo "Counts the number of current active sandboxes in the platform."
+  echo " "
+	echo "Usage: get-active-sandbox-and-space-count [START_DATE] [END_DATE]"
+  echo " "
+  echo -e "  START_DATE: \t A date to limit finding active sandboxes at or after the provided date in YYYY-MM-DD format."
```
bengerman13, line 19: the language is inconsistent, so I'm not sure the intended behavior here. Are we looking for:
- sandboxes created between the dates
- sandboxes that were active at any time between the dates

ccostino: Ah good catch - the script is doing a slightly different thing if a date is supplied, so let me clarify and see if that helps.  It may also be easier/better to just split the behavior apart into separate scripts.

All told, on the one hand have the ability to count all current active sandboxes.  Supplying dates is somewhat immaterial because sandboxes can only be active for 90 days at most and are then cleared out.

When you do supply dates though, this checks for how many spaces were created in sandbox organizations during that time period (or at least limited to on or after a date, or before and on a date if only one is supplied).

ccostino: @bengerman13 this should be a lot better now.  I think we can still have everything in here but the language, instructions, and intent should be much clearer!

bengerman13: the updated comment helps, but I think the help message is still unclear

ccostino: I'll tweak it further, thanks!

ccostino: Finally got the changes committed, this should be a lot better now!

bengerman13 (approved):

PR #200, opened 2022-02-04
Add support for an origin flag

ccostino:
This changeset adds support for accepting an origin flag to use when setting roles for a new organization. By default the flag is not used, but if the argument is present it will be passed in as a flag for the `set-org-role` and `set-space-role` commands.

- Add support for a `MANAGER_ORIGIN` argument
- Add a check to pass in the `--origin` flag for `cf set-org-role` and `cf set-space-role`

## Security considerations:
- None; the script can only be run by a platform operator, and adding the origin flag just lets CF know how to set the correct user

### DIFF
--- a/cf-create-org.sh
+++ b/cf-create-org.sh
@@ -4,7 +4,7 @@
 set -x
 
 if [ "$#" -lt 5 ]; then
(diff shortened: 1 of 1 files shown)


cf-create-org.sh
```diff
@@ -16,6 +16,13 @@ POP_END=$5
 MANAGER=$6
 
 MEMORY="${7:-4G}"
+MANAGER_ORIGIN="${8:-}"
```
bengerman13, line 19: This means you have to set memory any time you want to set manager origin

ccostino: Took the default value out for `MEMORY`.

PR #201, opened 2022-02-14
Fix create-org script

ccostino:
This changeset fixes an issue with the create-org script and the recently added `--origin` flag

- Fix handling of script arguments default values for the `--origin` flag

## security considerations:
- None


mheadd (approved):
👍

PR #213, opened 2022-05-06
Fix detection of 2.17.0 logstash

pburkholder:
-
- paths for 2.17.0 like  `/var/vcap/data/packages/logstash/801384be86dc1e096c8609c1ec9b8f7434538f6b/logstash-core/lib/jars/log4j-core-2.17.0.jar` are OK
-

## security considerations
1) This PR doesn't reveal anything senstive. a) Logstash is not susceptible and b) it's already OSS that we use logstash and 2) this improves security by removing false positive noise.


mheadd (dismissed):
👍

pburkholder:
No longer needed

PR #215, opened 2022-06-01
Update cloudy  bunch for June 2022

pburkholder:
-
- Alphabetize folks
- Updates for roster changes
- Better randomization logic

Now that the team is larger, simple randomization of the list makes for big swings in frequency, e.g., one person will appear 14 times, someone else only 7.

I changed the logic to randomize the list, then sample it every N, with the function `bunch`, then 
called the function twice.

This gives a consistent distribution of names.

## security considerations
Leaks internal emoji handles of team members. This probably less of an issue than our GitHub membership already. No impact.

### DIFF
--- a/cloudybunch.py
+++ b/cloudybunch.py
@@ -14,7 +14,7 @@ team = [
     ":david-corwin:",
     ":ephraim:",
     ":kelley:",
-    ":markboyd:",
+    ":mark-boyd:",
     ":markheadd:",
     ":melanie:",
     ":peterb:",


rcgottlieb (approved):

cloudybunch.py
```diff
@@ -3,23 +3,34 @@
...
+    ":ben-berry:",
+    ":mogul:",
+    ":brian-hurst:",
+    ":chiaka:",
     ":chris-mcgowan:",
+    ":david-anderson:",
+    ":david-corwin:",
+    ":ephraim:",
     ":kelley:",
+    ":markboyd:",
```
markdboyd, line 17:
```suggestion
    ":mark-boyd:",
```

PR #218, opened 2022-06-29
Write PGP public key to S3

markdboyd:

- Add script to write PGP public key used for commit signing to an S3 bucket

## security considerations

Public keys should be safe to share anyway, but this script shouldn't be introducing additional exposure regardless


rbogle (approved):
Awesome! Thanks @markdboyd !!

PR #219, opened 2022-06-30
Script to identify git resources without commit signing required

markdboyd:

- Adds script to identify `type: git` resources for cloud-gov/18F repos used in Concourse pipelines which do not have commit signing configured via `commit_verification_keys` 

## Questions

- Should this script also scan archived pipelines?

## security considerations

This script ultimately helps us improve security by identifying git resources in Concourse pipelines that don't verify commit signing so that we can fix them


markdboyd:
Question: should this script also scan archived pipelines?

rbogle (approved):

PR #220, opened 2022-07-13
add groups with permissions to manage routing for CF admins

markdboyd:

- add groups with permissions to manage routing for CF admins

## security considerations

Gives CF admins permission to manage routing for CF, but that seems desirable?


apburnes (approved):
Are we going to have to rerun this for ourselves to update our permissions?

markdboyd:
@apburnes Yes we would

PR #224, opened 2022-08-15
Add script to get memory for all apps in an org

markdboyd:

- Add script to get memory for all apps in an org. Helpful for debugging issues related to customers reaching their org or space quota memory limits

## security considerations

None


markdboyd:
https://plugins.cloudfoundry.org/#Usage%20Report already exists and is more fully featured than this script

PR #226, opened 2022-08-25
Add script for reformatting credentials for Credhub import

jameshochadel:
- Add script for reformatting credentials for Credhub import

## security considerations
None. This change is related to https://github.com/cloud-gov/product/issues/2081, which improves the security of the platform.

### DIFF
--- a/credhub-format.sh
+++ b/credhub-format.sh
@@ -4,7 +4,7 @@ set -eo pipefail
 # yq operates on yaml documents using a jq-like syntax.
 # For jq program syntax, see: https://stedolan.github.io/jq/manual/
 
-if [ "$#" -ne $2 ]; then
+if [ "$#" -ne 2 ]; then
   echo "
 Usage: ./credhub-format.sh <source-file> <pipeline-name>
 Example: ./credhub-format.sh credentials.yml deploy-something > output.yml


credhub-format.sh
```diff
@@ -0,0 +1,64 @@
...
+# yq operates on yaml documents using a jq-like syntax.
+# For jq program syntax, see: https://stedolan.github.io/jq/manual/
+
+if [ -z $2 ]; then
+  echo "
+Usage: ./credhub-format.sh <source-file> <pipeline-name>
+Example: ./credhub-format.sh credentials.yml deploy-something > output.yml
+
+Read from a credential file and reformat the contents to a JSON format that CredHub
+can import, excluding any keys that do not appear in ci/pipeline.yml.
```
markdboyd, line 13: So is the main point of the script that it excludes credentials which aren't used in the pipeline?

credhub-format.sh
```diff
@@ -0,0 +1,64 @@
...
+  exit 1
+fi
+
+# check silently if yq is installed.
+if ! which -s yq; then
+  echo "yq is required to run this script." >&2
+  exit 1
+fi
+
+if [ ! -f ci/pipeline.yml ]; then
```
markdboyd, line 27: Not sure if it's better, but if we don't want to be dependent on the `ci/pipeline.yml` path convention, you could use `fly -t ci get-pipeline -p $pipelinename` to download the currently pipeline definition. Though I'm not sure if that has the credhub values already interpolated from variables or not

rcgottlieb (approved):

credhub-format.sh
jameshochadel: The main point is to convert from our existing yaml credential files format:

```yaml
key: value
```

to the format Credhub's `import` command expects, JSON that looks like: 

```json
{
  credentials: [
    {
      "name": /concourse/main/repo-name/key,
      "type": "value",
      "value": <value>,
    },
    ...
  ]
}
```

The exclusion of unused keys is a bonus to make sure we're not carrying cruft into the new system.

credhub-format.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+set -eo pipefail
+
+# yq operates on yaml documents using a jq-like syntax.
+# For jq program syntax, see: https://stedolan.github.io/jq/manual/
+
+if [ -z $2 ]; then
```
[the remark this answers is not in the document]
jameshochadel: Since the args are positional, if you pass only zero or one args, this will evaluate to false. And if you pass two, you're golden. That said, checking the count is probably more clear and conventional. I'll switch to that.

credhub-format.sh
jameshochadel: I like that idea, but unfortunately the variables are already interpolated as you guessed.

markdboyd (approved):

PR #228, opened 2022-10-05
Service info

rbogle:
- adds a script which provided the name of a service ( and org/space) can query aws for the info on that resource.
- outputs the info in json format combining cf info and aws info
-

## security considerations
None, temporary credentials are created and destroyed and not used in output.


rcgottlieb (approved):
You might want to change the comment  "get the rds db_instance info" to whatever the query is doing in the body of each of them otherwise I'm approving as it looks good for the actual code.

PR #231, opened 2022-10-31
fix create org script

markdboyd:

Initially running this script I got some errors because `$ORIGIN_FLAG` was equal to `""`, so extra `''` were showing up in the `cf` commands and causing them to fail. The BASH substitution syntax I've added should fix the errors.

## security considerations

None, just fixing the commands in this script


jameshochadel (approved):
Nice catch. One day I'll learn all of the bash substitution syntax...

PR #237, opened 2023-01-20
Update Pages production domain services to point CDN origin to the pages proxy 

svenaas:
**WORK IN PROGRESS** towards cloud-gov/pages-core#4015

Adds `pages/update-cdn-origins.sh` which retrieves domains from the `domain` table, updates the CDN origin as necessary via `cf update-service` to update the CDN origin from Federalist to cloud.gov Pages standard, and updates the `domain` table with the new origin.

The script is expected to be run from a jumpbox environment from which it can access the sites space in Cloud Foundry and the relevant database.

It requires six environment variables:
- `CF_API_URI`
- `CF_USERNAME`
- `CF_PASSWORD`
- `CF_ORG`
- `CF_SPACE`
- `DB_URI`

And it takes one optional command line argument: 
```
update-cdn-origins.sh [max-domains]
```
`max-domains` allows execution to be limited to specific number of domains rather than processing all unprocessed domains in the `domain` table.

- Adds `pages/update-cdn-origins.sh`

## security considerations
No scripts are introduced into the running codebase. This is a script for use by administrators under specific circumstances.

### DIFF
--- a/pages/update-cdn-origins.sh
+++ b/pages/update-cdn-origins.sh
@@ -1,0 +1,103 @@
+#! /bin/bash
+
+set -e
+
+display_usage() {
+  echo -e "\nUsage: $0 input.csv [max-rows] [offset]\n"
+  echo -e "  max-rows: process at most this number of input rows (default: unlimited)\n"
+  echo -e "  offset:   skip an initial set of input rows (default: 0)\n"
+}
+
+if [  $# -le 0 ]
+then
+  display_usage
+  exit 1
+fi
+
+if [[ ( $@ == "--help") ||  $@ == "-h" ]]
+then
+  display_usage
+  exit 0
+fi
+
+# Decision made to run this from a different environment. Hence:
+# TODO: Eliminate input file and instead query DB from within the script
+# TODO: Eliminate SQL output and instead update DB from within primary loop
+
+# Expected input file is a comma-seperated-value export of the serviceName and
+# origin columns for provisioned domains in the domains table in the core DB.
+#
+# The expected format can be produced from the psql command line as follows:
+#
+# \copy (select "serviceName",origin from domain where state='provisioned') to '/output/path/for/domains.csv' CSV HEADER;
+domains_file=$1
+
+max_rows=${2:--1}
+
+offset=${3:-0}
+
+if [[ ! -r $domains_file ]]
+then
+  echo "Input file $domains_file does not exist or is not readable"
+  exit 1
+fi
+
+# QUESTION: Should this type of script test for CF session and/or verify targeted org/space?
+
+# QUESTION: Does wait_for_service_instance() require the following?
+#
+# CF Auth
+# cf api "${CF_API_URL}"
+# (set +x; cf auth "${CF_USERNAME}" "${CF_PASSWORD}")
+
+# Waiting for service instance to finish being processed.
+wait_for_service_instance() {
+  local service_name=$1
+  local guid=$(cf service --guid $service_name)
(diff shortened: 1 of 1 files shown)


svenaas:
@drewbo @apburnes The DB query code in this has been tested from the production jumpbox, but the `cf update-service` and DB update code have not been. Obviously I don't want to try them in production before we try them staging, and that will require setting up one or more domains there in a suitable state for this to modify them. Testing has also been piecemeal, not full script, as its the merging of this PR that will make the full code available on the jumpbox.

apburnes:
@svenaas provisioning federalist-test-site.18f.gov to the staging env.

svenaas:
@apburnes I see it there, but with a bucket path in `sites.pages-staging.cloud.gov`. 

As currently implemented my script is querying for provisioned domains with `app.cloud.gov` origins and updating them to `sites.pages.cloud.gov`. 

So I've got two questions here: 
1. Can federalist-test-site.18f.gov be put into a working pre-Pages state for testing purposes?
2. Should the script be generalized to look for pre-Pages staging URL patterns and if so, what does that pattern look like?

apburnes:
We have no current infrastructure to deploy "federalist" endpoints but it is an opportunity to test the script to make sure it doesn't update a `*.sites.<pages-env>.cloud.gov` accidentally. Also, you'll have to update the service and database by hand to run the script but it will be good practice.

apburnes:
Feel free to update the service's origin and database entry in staging to help with your testing.

svenaas:
That part works from my local. I think I'm going to run into a snag from the jumpbox; determining that next.

svenaas:
The code in this script works against a test case in staging. It will be nice to run it again there via cg-scripts after this is merged, though, to be sure nothing is different when executed that way (and that I didn't miss additional spots). 

The `wait_for_service_instance` call runs for a good long time — I exceeded the jumpbox idle timeout before it finished — and we need to choose a strategy for dealing with that so that it doesn't interrupt execution of the rest of the script.

Do y'all think there should be more error handling added (in the main loop or elsewhere)? As it stands I'm outputting progress notes and relying on `set -e` to pull the brakes, which is arguably sufficient for the purpose.

svenaas:
Running this for a while on a jumpbox to see whether it's enough to keep the connection open, or whether it still times out: 
```
while [ "true" == "true" ]; do
  sleep 60
  echo "."
  done
}
```

svenaas:
Alas, it still times out.

pages/update-cdn-origins.sh
```diff
@@ -0,0 +1,81 @@
...
+
+  while [ "$status" == "in progress" ]; do
+    sleep 60
+    status=$(cf curl /v2/service_instances/${guid} | jq -r '.entity.last_operation.state')
+  done
+}
+
+# Find domains that need to be updated and iterate through them
+domains_processed=0
+query="select \"serviceName\",origin from domain where state='provisioned' and origin like '%app.cloud.gov';"
```
apburnes, line 46: Should we also pull out the domain's unique id to use on the update sql so we know we are updating the right instance?

apburnes (commented):
Looks good. Just a question on the sql query.

pages/update-cdn-origins.sh
svenaas: Yes, that's a good suggestion. I hadn't noticed there could be duplicate serviceNames. Thanks.

apburnes (approved):
Awesome. Once you merge this, login to concourse and rerun the jumpbox CI job so it will be able to pick up the new script.

svenaas:
OK, I think I should be able to figure that out. I'll do it after my training session when I'm giving it my full attention, though.

PR #278, opened 2023-10-11
Update aws-list-unused-rds-databases.py

cweibel:
- Retrieve org name and space name from CF
- Created "action" column with business logic on how to handle a particular entry
- Fix to how db connections are calculated, switched to pulling stats by day and summing the results as to not miss any of the history by attempting to match period length to start/end time.

## security considerations
None, relies on aws-vault and cf oauth-token for access


aws-list-unused-rds-databases.py
markdboyd, line 1: My only nit is that we should put this in the `aws` folder so that we can start organizing the scripts in this repo better

cweibel: Sounds good, fixing

markdboyd (approved):

PR #292, opened 2023-12-14
Update CF role scripts

markdboyd:

- Update cf-get-user-roles.sh to return a list of roles with the names of orgs/spaces rather than GUIDs. The output format is no longer CSV.

    Before:
    
    ```shell
    role                  org_uuid                                space_uuid
    "organization_user"  "abc-123"  "null"
    "organization_user"  "def-456"  "null"
    "space_manager"      "null"                                  "ghi-789"
    ```
    
    After:
    
    ```shell
    organization_user: org-1
    organization_user: org-1
    space_manager: space1
    space_developer: space1
    ```
- Move helper scripts for querying CF API to `lib/cf.sh`
- Refactor scripts to use shared CF API helpers
- Add .shellcheckrc to suppress https://www.shellcheck.net/wiki/SC1091
- Update `strip-user-org-and-space-roles.sh` to suppress errors on older BASH versions

## security considerations

None, just refactoring and improving helper scripts

### DIFF
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,0 +1,1 @@
+external-sources=true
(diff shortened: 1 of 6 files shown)


jameshochadel (approved):
Officially my favorite PR of the year. THANK YOU!

rcgottlieb (approved):

PR #293, opened 2024-01-03
Add defect dojo daemon for parse-nessus

pburkholder:
- Updates summary output to match format from CVSSv2
- Moves CVSSv3 logic to parsing step, not output step
- Add defect dojo daemon regex for deamon check
- Use "Medium" not "Moderate" risk

## security considerations
Safe. Approved new services.


wz-gsa (approved):
Addresses issues as discussed

PR #299, opened 2024-02-28
Prometheus scope

rcgottlieb:
- Adding a new script to add users to a new group for grafana access
-
-

## security considerations
No

### DIFF
--- a/make-prometheus-support.sh
+++ b/make-prometheus-support.sh
@@ -5,7 +5,7 @@ set -e
 if [ "$#" -lt 1 ]; then
   echo
   echo "Usage:"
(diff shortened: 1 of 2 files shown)


make-prometheus-support.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+set -e
+
+if [ "$#" -lt 1 ]; then
+  echo
+  echo "Usage:"
+  echo "  $ ./make-ops-viewer.sh [-r] <EMAIL_ADDRESS>"
```
jameshochadel, line 8: I think this could use an update; otherwise looks good!

rcgottlieb: Good catch! Thank you! Fixed with next commit.

jameshochadel (approved):

PR #302, opened 2024-03-07
add log filename finder script

bengerman13:
- add log filename finder script

## security considerations
None - this script only references AWS general knowledge

### DIFF
--- a/logs-from-timestamp.py
+++ b/logs-from-timestamp.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python3
+"""
+Given a unix timestamp, list the prefix for files that might contain relevant logs
+usage:
+$ python3 logs-from-timestamp.py 1709655649692
+s3://bucket/AWSlogs/[...]domains-2[...]20240301T0105
+s3://bucket/AWSlogs/[...]domains-1[...]20240301T0105
+s3://bucket/AWSlogs/[...]domains-0[...]20240301T0105
+"""
+
+import sys
+import datetime
+
+# hours to offset source time from utc
+OFFSET_TIME = -5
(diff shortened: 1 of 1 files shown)


logs-from-timestamp.py
```diff
@@ -0,0 +1,49 @@
...
+	
+
+def floor_to_five_minutes(dt):
+	td = datetime.timedelta(minutes=(dt.minute % 5))
+	return dt - td
+
+
+def filename_from_datetime(target_dt, alb_name):
+	s = f"s3://{BUCKET}/AWSLogs/{ACCOUNT_ID}/elasticloadbalancing/{REGION}" + \
+		f"{ACCOUNT_ID}_elasticloadbalancing_{REGION}_app.{alb_name}_{target_dt.strftime('%Y%m%dT%H%M')}"
```
rcgottlieb, line 43: On second glance looking at the actual bucket in S3, there is a / missing after {REGION}, and there is a four digit year following REGION, 2 digit month, 2 digit day missing as well, so assuming YEAR, MONTH, DAY variables, it should probably look something like this:
s = f"s3://{BUCKET}/AWSLogs/{ACCOUNT_ID}/elasticloadbalancing/{REGION}/{YEAR}/{MONTH}/{DAY}/" + \
		f"{ACCOUNT_ID}_elasticloadbalancing_{REGION}_app.{alb_name}_{target_dt.strftime('%Y%m%dT%H%M')}"

rcgottlieb (changes requested):

download_elb_logs.py
markdboyd, line 1: Could we put all of these scripts in the `aws` directory?

PR #306, opened 2024-03-20
add script for creating users in UAA

markdboyd:

- add script for creating users in UAA

## security considerations

None

### DIFF
--- a/uaa/add-user.sh
+++ b/uaa/add-user.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+set -e
+
+if [ "$#" -lt 1 ]; then
+  echo
+  echo "Usage:"
+  echo "  $ ./add-user.sh <EMAIL_ADDRESS>"
+  echo
+  exit 1;
+fi
+
+if ! hash uaac 2>/dev/null; then
+  echo -n "uaac must be installed"
+  exit 1
+fi
+
+EMAIL=$1
+ORIGIN=$(awk -F '@' '{print $2}' <<< "$EMAIL")
+
+echo -n "Adding user ${EMAIL}... "
+uaac curl \
+  -X POST /Users \
+  -H "If-Match:*" \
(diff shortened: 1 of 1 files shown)


uaa/add-user.sh
```diff
@@ -0,0 +1,28 @@
...
+  exit 1;
+fi
+
+if ! hash uaac 2>/dev/null; then
+  echo -n "uaac must be installed"
+  exit 1
+fi
+
+EMAIL=$1
+ORIGIN=$(awk -F '@' '{print $2}' <<< "$EMAIL")
```
markdboyd, line 19: Is it too clever to determine the origin from the email? it seemed safe to me and worked in testing

jameshochadel: Hm, maybe just note that behavior in the usage text? I can see someone wanting to create a user from the cloud.gov origin with an agency email.

uaa/add-user.sh
```diff
@@ -0,0 +1,28 @@
...
+
+echo -n "Adding user ${EMAIL}... "
+uaac curl \
+  -X POST /Users \
+  -H "If-Match:*" \
+  -H "Accept:application/json" \
+  -H "Content-Type:application/json" \
+  -d \{\"userName\":\""${EMAIL}"\",\"emails\":[\{\"value\":\""${EMAIL}"\"\}],\"active\":true,\"verified\":true,\"origin\":\""${ORIGIN}"\"\}
```
jameshochadel, line 27: Quick question — why use `uaac curl` instead of one of the built-in commands, like `uaac user add`? If there's a good reason, can you add a comment documenting it here?

markdboyd: Good question. Based on a test I just ran, it looks like `uaac user add` requires interactivity to specify the username, so wouldn't be helpful for automation, whereas the `uaac curl` works without the interactivity

jameshochadel (approved):

PR #308, opened 2024-03-27
fixed additional logic for defect dojo daemon

wz-gsa:
- improved logic to always catch defectdojo daemon
-
-

## security considerations
none

### DIFF
--- a/parse-nessus-xml.py
+++ b/parse-nessus-xml.py
@@ -94,7 +94,8 @@ DAEMONS = """
     cc.uploader
     cf_exporter
     concourse
-    discovery-registrar
+    discovery-registra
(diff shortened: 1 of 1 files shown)


parse-nessus-xml.py
```diff
@@ -94,7 +94,8 @@ def remediation_plan(vuln):
     cc.uploader
     cf_exporter
     concourse
-    discovery-registrar
+    discovery-registra
```
dandersonsw, line 97: Should the 'r' be deleted here?

wz-gsa: fixed

dandersonsw (approved):

PR #320, opened 2024-06-12
made nessus subfolder and added script for fixing container license

wz-gsa:
- added nessus folder
- added fix_nessus_license.sh
- added readme_fix_nessus_license.md

## security considerations
Improves security by ensuring plugins remain updated

### DIFF
--- a/github/add_security_md_to_json_list.py
+++ b/github/add_security_md_to_json_list.py
@@ -157,7 +157,7 @@ def add_commit_push_security_md(repo_path, branch_name):
 
 # Create a pull request
 def create_pull_request(repo_path, branch_name, default_branch):
-    """Create a pull request for the branch, attempt to add reviewers, and assign 'wz-gsa'."""
+    ""f"Create a pull request for the branch, attempt to add reviewers, and assign '{ASSIGNEE}'."""
     original_dir = os.getcwd()  # Save the current directory
     try:
         os.chdir(repo_path)  # Change to the repo's directory
@@ -195,7 +195,7 @@ def create_pull_request(repo_path, branch_name, default_branch):
 
         if (
             "Reviewers could not be requested" in result
-            or "Assignee could not be added" in result
+            or f"{ASSIGNEE} could not be added" in result
         ):
             logging.warning(
                 "Attempting to add 'cloud-gov-pages-operations' as a fallback reviewer."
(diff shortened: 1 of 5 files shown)


github/add_security_md_to_json_list.py
```diff
@@ -157,7 +157,7 @@ def add_commit_push_security_md(repo_path, branch_name):
 
 # Create a pull request
 def create_pull_request(repo_path, branch_name, default_branch):
-    """Create a pull request for the branch, attempt to add reviewers, and assign 'wz-gsa'."""
+    ""f"Create a pull request for the branch, attempt to add reviewers, and assign '{ASSIGNEE}'."""
```
bengerman13, line 160: where is `ASSIGNEE` being set?

bengerman13, line 160: With the `f` in the middle of the quotes, you no longer have a triple-quoted string, you have three strings being silently concatenated.

This should probably be either:
`f"""Create a pull request for the branch, attempt to add reviewers, and assign '{ASSIGNEE}'."""`

or:
`f"Create a pull request for the branch, attempt to add reviewers, and assign '{ASSIGNEE}'."` (since the triple-quotes aren't actually necessary here)

github/parse_repo_contents_into_txt.py
bengerman13, line 1: what's this for? doesn't seem to be mentioned in the PR description or READMEs.

nessus/fix_nessus_license.sh
markdboyd, line 1: No idea if this is relevant, but this is the script on the Nessus BOSH release the determines if the license is valid and alerts us:

https://github.com/cloud-gov/cg-nessus-manager-boshrelease/blob/main/jobs/nessus-manager/templates/bin/health.sh

Wondering if any of these improvements can go to that script so we don't get alerts in the first place?

github/parse_repo_contents_into_txt.py
wz-gsa: it's addition was unintentional on my part. updating my commit and reverting this to draft

nessus/fix_nessus_license.sh
wz-gsa: This was based off of that, and may serve as a replacement, but that bosh script was not fixing the underlying issue and required manual intervention yesterday.

markdboyd: I'm not sure I understand. It looks like this script does the same thing as that BOSH script: detecting if the license is outdated/plugins are outdated and setting Prometheus metrics. So if the changes in this script make things work "correctly" so that we don't get false positives on the alerts, why wouldn't we just replace the BOSH script with this one?

wz-gsa: We probably can. I wanted change control on it and to make this available and use this a few times before moving it there to ensure I tested it thoroughly before shipping it in the bosh release.

dandersonsw:
Closing this after merging in the change to the bosh release: https://github.com/cloud-gov/nessus-manager-boshrelease/pull/51

PR #324, opened 2024-07-31
make query to delete user TOTP do a case-insensitive

markdboyd:

- make query to delete user TOTP do a case-insensitive by using PostgreSQL `lower` function

## security considerations

None, just making this script easier to use so that a username will be deleted regardless of what case the username is entered as


markdboyd:
Nevermind, this won't work because you can have the same email with different UAA accounts based on case sensitivity

PR #334, opened 2024-12-05
add check for whether database exists before attempting to get tags

markdboyd:

- Update script to detect if database exists before attempting to get tags for it

## security considerations

None, just improving script


pburkholder (approved):
thanks for putting SUCCESS _after_ it succeeds!

PR #338, opened 2025-03-13
refactor script for getting services

markdboyd:

- refactor script for getting services to filter by organization or service offering name
- rename script for clarity

## security considerations

None - script contents are not sensitive


pburkholder:
When I run this without any options, `./get-services.sh`, it gets every service in every org, and I can't kill it with ^c or ^z.

That may be feature -- but it feels like a bug.

cloudfoundry/get-services.sh
```diff
@@ -0,0 +1,84 @@
...
+}
+
+while getopts ":hs:o:" opt; do
+  case ${opt} in
+    h )
+      usage
+      exit 0
+      ;;
+    o )
+      ORGANIZATION_GUID=$OPTARG
```
pburkholder, line 26: How about making it `-o org_name` then here run `ORGANIZATION_GUID=$(cf org --guid $OPTARG)`

markdboyd: Good idea! Changed: 03aeb2f68d22a81dd989ea18d1c6b6f8d08c8380

markdboyd:
@pburkholder Good catch! I updated the script that either the org name or service offering name must be provided as options to the script, so that by default it doesn't try to list all services on the platform

pburkholder (approved):

PR #347, opened 2025-04-09
modified one script to move from v2 to v3

rcgottlieb:
- updated the only platform script to v3
-
-

## security considerations
None


cweibel (approved):
Appears to be the correct path when run manually.

LGTM

PR #350, opened 2025-04-22
add script to update plan visibility

markdboyd:

- add script to update plan visibility. It is especially useful when you need to change a service plan's visibility from `all` to a specific set of orgs

## security considerations

None. The script contents are not sensitive and it can only be run by platform admins


rcgottlieb (approved):

pburkholder (changes requested):
- I learned from you the `jq` trick to url-encode user input. Use that here?
- Add bash `set -euo pipefail` for "standard" safety check

PR #353, opened 2025-04-23
update create org script to set label of org-type=customer

markdboyd:

- update create org script to set label of org-type=customer

Having a consistent label for customer orgs would allow us to [easily query for them using the CF API](https://v3-apidocs.cloudfoundry.org/version/3.188.0/index.html#labels-and-selectors) rather than relying on `grep` to exclude orgs by name, which is unreliable and brittle over time. We could update the script in https://github.com/cloud-gov/cg-scripts/pull/352 to use the label when querying

## security considerations

None. The script contents and these changes are not sensitive


spgreenberg (approved):

pburkholder:
Trouble is, I've used this script to create orgs (like cloud-gov-training) for non-customer purposes to make sure it's set up with reasonable defaults. How do we prevent that from happening?

PR #376, opened 2025-12-31
Script up process for enabling PGAudit

pburkholder:

- This is the first step, enabling
- The configuring may be do-able by the customer via SQL scirpts
-

## security considerations
[Note the any security considerations here, or make note of why there are none]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ my*.yml
 *.xlsx
 .vscode
 venv
+.venv
 **/__pycache__
 *.txt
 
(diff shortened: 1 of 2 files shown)


zjrgov (commented):
Apologies if I've missed it here or in the discussions otherwise but I was wondering: what sort of credentials are you using with the AWS CLI for this?

markdboyd:
Ultimately, I think this may be better handled as a feature of the AWS broker, so closing for now

PR #390, opened 2026-03-23
Adding script to bulk reset aide

ChrisMcGowan:
- Bulk script that finds all deployments on the targeted BOSH director, sorts instance groups, and runs aide reset

## security considerations
n/a


cweibel (approved):
Good for when the bosh director upgrades

PR #392, opened 2026-04-07
Update log export for OpenSearch

pburkholder:
-
- Runs on OpenSearch Manager
- Uses TLS
- Changes exit criteria from d/lo loop
- Compresses final product
- Creates just one .CSV file

## security considerations
uses TLS, improvement.

### DIFF
--- a/logs/get-csv-logs.sh
+++ b/logs/get-csv-logs.sh
@@ -1,39 +1,44 @@
 #!/bin/bash
 
+set -euo pipefail
+
 function usage {
   echo -e "
-  ./$( basename "$0" ) <IP_OR_HOSTNAME> <SEARCH_JSON_FILE> <'field1,field2'>
+  ./$( basename "$0" ) <SEARCH_JSON_FILE> [CSV_FIELDS]
 
   Use the Elasticsearch/Opensearch scroll API to query logs for bulk export
-  See full docs at: https://github.com/cloud-gov/internal-docs/blob/main/docs/runbooks/Logging/bulk-data-export.md
+  See full docs at: https://workshop.cloud.gov/cloud-gov/internal-docs/-/blob/main/docs/runbooks/Logging/bulk-data-export.md
 
-  2026-04-06: Updated to use TLS key and cert for authentication
+  With Opensearch using TLS, run this from an an OpenSearch Manager node as the vcap user.
+
+  Then use bosh ssh to copy the file to the jumpbox, and from there to your local machine.
 
   Examples:
-    ./$( basename "$0" ) 127.0.0.1 search.json '."@timestamp",.rtr.path,.rtr.status'
+    ./$( basename "$0" ) search.json
   "
   exit
 }
 
-if [ -z "$1" ]; then
-  echo "IP/hostname for Elasticsearch/Opensearch cluster is required as first argument"
-  usage
-fi
+CSV_FIELDS='."@timestamp",."@raw"'
+ES_URL="localhost:9200"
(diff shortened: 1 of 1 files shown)


logs/get-csv-logs.sh
```diff
@@ -1,37 +1,44 @@
...
 
-if [ -z "$1" ]; then
-  echo "IP/hostname for Elasticsearch/Opensearch cluster is required as first argument"
-  usage
-fi
+CSV_FIELDS='."@timestamp",."@raw"'
+ES_URL="localhost:9200"
+CERT_PATH="/var/vcap/jobs/opensearch/config/ssl/opensearch-admin.crt"
+KEY_PATH="/var/vcap/jobs/opensearch/config/ssl/opensearch-admin.key"
+CA_PATH="/var/vcap/jobs/opensearch/config/ssl/opensearch.ca"
```
markdboyd: What do you think about making these three variables something that you have to set on the command line before running this script? I have some hesitation about hardcoding these paths into this script, but at the same time I can't think of a very compelling reason why.

JasonTheMain:
We need a way or a procedure stating to clean up the document on the vm

JasonTheMain:
> We need a way or a procedure stating to clean up the document on the vm

I think this script needs to store them in the /var/vcap/store. that way if its forgotten we aren't putting a manager in danger

logs/get-csv-logs.sh
```diff
@@ -1,37 +1,44 @@
...
+OUTPUT_FILE="logs_export.csv"
```
ChrisMcGowan: Please change this to be somewhere under `/var/vcap/store` as currently this script dumps the CSV file under `/home/vcap` which is the root OS disk and could cause the VM to crash if root runs out of disk

logs/get-csv-logs.sh
pburkholder: I'll update to match how we do it https://github.com/cloud-gov/opensearch-boshrelease/blob/main/jobs/opensearch/templates/bin/drain.erb#L13-L15, so it seems less explicit or leaky.

logs/get-csv-logs.sh
pburkholder: Got it. I've also added a `/bin/rm step` to the documentation in internal-docs.

markdboyd (dismissed):

logs/get-csv-logs.sh
```diff
@@ -1,37 +1,46 @@
...
-  usage
-fi
+JOB_NAME=opensearch
+JOB_DIR=/var/vcap/jobs/$JOB_NAME
+CSV_FIELDS='."@timestamp",."@raw"'
+ES_URL="localhost:9200"
+CERT_PATH="${JOB_DIR}/config/ssl/opensearch-admin.crt"
+KEY_PATH="${JOB_DIR}/config/ssl/opensearch-admin.key"
+CA_PATH="${JOB_DIR}/config/ssl/opensearch.ca"
+OUTPUT_FILE="/var/vcap/store/logs_export.csv"
```
markdboyd:
```suggestion
OUTPUT_FILE="/var/vcap/store/tmp/logs_export.csv"
```

I think it will be nice to have a the file in a folder which clearly indicates it can be deleted/cleaned up in case someone forgets to clean it up manually

pburkholder: Good point. I've updated to create an `export` dir, and the docs MR has that as well: https://workshop.cloud.gov/cloud-gov/internal-docs/-/merge_requests/451/

PR #394, opened 2026-06-30
Remove space roles before org role

pburkholder:

- Fix: 
```
{"errors":[{"detail":"Cannot delete organization_user role while user has roles in spaces in that organization.","title":"CF-UnprocessableEntity","code":10008}]}
```
- Update usage
- Add a sleep before removing org roles (it failed the first time probably because a change hadn't propagated)
- Echo the commands to provide feedback on what's happening
- 

## security considerations

Safe: strips user roles, with some constraints to prevent out-of-bound loops

### DIFF
--- a/cloudfoundry/strip-user-org-and-space-roles.sh
+++ b/cloudfoundry/strip-user-org-and-space-roles.sh
@@ -3,6 +3,9 @@
 set -euo pipefail
 shopt -s inherit_errexit || true
 
+MAX_SPACE_ROLES=40
+MAX_ORG_ROLES=8
+
 main() {
   [[ $# -eq 2 ]] || usage "Expected two arguments, got $#"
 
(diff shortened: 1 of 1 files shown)


cloudfoundry/strip-user-org-and-space-roles.sh
```diff
@@ -16,24 +19,43 @@ main() {
...
+    done
+  fi
 
-  # get all user roles for the org, including organization_user
-  for role_guid in $(cf curl "/v3/roles?user_guids=$USER_GUID&organization_guids=$ORGANIZATION_GUID&per_page=5000" | jq -r '.resources[].guid'); do
+  # now remove all org-level roles for the user, including organization_user
+  for role_guid in $(cf curl "/v3/roles?user_guids=$USER_GUID&organization_guids=$ORGANIZATION_GUID&per_page=$MAX_ORG_ROLES" | jq -r '.resources[].guid'); do
+    echo removing org role with: cf curl -X DELETE "/v3/roles/$role_guid"
+    # Let the above role removals propagate (sleep 1 was insufficent)
+    sleep 5
```
markdboyd, line 44: Why is the sleep necessary at all?

pburkholder: Because the first time I ran the updated script if failed with:

"Cannot delete organization_user role while user has roles in spaces in that organization."

At that time I had an echo and 1s sleep before every DELETE so I could kill the script if it went off the rails. So I know a 1s sleep is insufficient for the space role deletion to fully propagate.

On the re-run the org role deletion succeeded.

So I know that
- running the org role deletion 1 second after an space role deletion fails
- running the org role deletion later works

and I figured maybe 5 seconds will work the next time.

We could put the org role deletion in a loop until it succeeds, but a sleep is simpler and time is not of the essence here.

markdboyd (approved):

PR #395, opened 2026-07-30
Cost estimator: Handle instances with missing tags

markdboyd:

- update cost estimator script to skip instances without an instance GUID tag and to log them

## security considerations

None


JasonTheMain (approved):
LGTM

PR #399, opened 2026-09-18
update script for getting org managers

markdboyd:

update script for getting org managers to:

- only include customer orgs
- exclude null usernames
- simplify the shell commands

## security considerations

None


JasonTheMain (approved):
Looks good to me

